PR #106, opened 2026-08-01
feat(manifest): add blacksmith CLI alternative

ViperJuice:
## Summary

Adds a `blacksmith` entry under `cli_alternatives` so agents discover the Blacksmith CLI instead of hunting for an MCP server that does not exist.

## Why cli_alternatives and not servers

**Blacksmith publishes no MCP server.** `[code]smith` is an MCP *client* — per [their integrations docs](https://docs.blacksmith.sh/codesmith/integrations), it consumes external servers ("Local MCP servers will be installed and run on Blacksmith infra in an isolated sandbox") rather than exposing one. Testbox integrates with agents through a **generated skill file**, not MCP. Their docs describe the CLI as *"agent-first ... the intended consumer is a coding agent, not a human"*, which is exactly what this section is for.

The reasoning is recorded in a comment above the entry so it is not re-researched later.

## How this was checked

| Check | Result |
|---|---|
| Full docs index (`llms.txt`, 30 pages) | No MCP-server page; only mention is *consuming* MCP |
| Public MCP Registry (`gateway.search_registry`) | No results |
| Advertised OpenAPI spec | `404 Asset not found` — no public REST API to wrap |
| `mcp.blacksmith.sh` | NXDOMAIN |
| `api.blacksmith.sh/mcp` | 404 |
| `app.blacksmith.sh/mcp` | Next.js `[org]` catch-all, not an endpoint |

## Community package rejected

`npm blacksmith-mcp` exists but is not manifest-worthy: it authenticates by **extracting the user's Blacksmith session cookie out of Chrome** (there is no public API to hold a token against), and has 3 stars / 11 commits / 34 weekly downloads.

## Test plan

- [x] `manifest.yaml` parses; `cli_alternatives` count 11 → 12
- [x] `uv run pytest tests/test_manifest.py` — 102 passed, 1 skipped
- [x] Matches the manifest-only shape of #102 (pangram): no version bump, no CHANGELOG entry

---
<a href="https://app.blacksmith.sh/Consiliency/codesmith/pmcp/pr/106"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a>

### DIFF
--- a/src/pmcp/manifest/manifest.yaml
+++ b/src/pmcp/manifest/manifest.yaml
@@ -93,6 +93,21 @@ cli_alternatives:
       - "terraform fmt -check"
       - "terraform providers"
 
+  # Blacksmith ships no MCP server: [code]smith is an MCP *client*, and Testbox
+  # integrates with agents through a generated skill file, not a server. The CLI
+  # is the supported agent surface ("agent-first ... the intended consumer is a
+  # coding agent"), so it belongs here rather than under servers:.
+  blacksmith:
+    keywords: [blacksmith, testbox, github actions, ci runner, ci analytics, flaky tests]
+    check_command: ["blacksmith", "--version"]
+    help_command: ["blacksmith", "--help"]
+    description: "Blacksmith CI runners and Testbox CLI"
+    examples:
+      - "blacksmith testbox status"
+      - "blacksmith testbox warmup <workflow>"
+      - "blacksmith testbox run --id <id> \"<command>\""
+    prefer_mcp_for: [github issues, pull requests, github api]
+
   curl:
     keywords: [curl, http, request, api, download]
     check_command: ["curl", "--version"]


ViperJuice:
## Self-review

Two things checked against the code rather than assumed:

**`examples` with placeholders — safe.** Every other `cli_alternatives` entry uses directly runnable examples (`git status --short`), while this one uses `<workflow>` / `<id>` placeholders. That is fine: `examples` are consumed only for text-match scoring in `matcher.py:133-136` and are never executed.

**`check_command: ["blacksmith", "--version"]` — unverified, low risk.** Blacksmith's [CLI reference](https://docs.blacksmith.sh/blacksmith-testbox/cli) documents only `testbox warmup|run|status|stop|init` and `auth login` — no global `--version` flag. The CLI is not installed on any host here, so I could not confirm it, and verifying would mean piping their install script to `sh`, which is out of scope for this change.

Impact if the flag is unsupported: none to detection. `check_cli` gates on `shutil.which` and explicitly treats a non-zero exit as still usable (`environment.py:82-84`) — only the version string is lost. `[name, "--version"]` is also the loader's own default (`loader.py:342`), so omitting the field would produce the identical command.

Worth a follow-up correction by anyone who has the CLI installed.

---

Unrelated but noticed while checking CI: this repo already runs a `[code]smith` check on PRs, so Blacksmith is already in use here.

PR #109, opened 2026-08-01
feat(manifest): add server_env overlay patch for existing servers

### LINKED ISSUE #105: Manifest entries support only one env var, blocking self-hosted server endpoints
## Problem

A manifest server entry can declare exactly one environment variable (`env_var`, with an optional namespaced `secret_key` for storage). Servers that need a second, non-secret variable — most commonly a base URL pointing at a self-hosted deployment — cannot be fully expressed in the manifest.

`build_install_child_env` (`src/pmcp/manifest/installer.py:509`) builds the child environment by resolving the credential for the single declared `env_var`:

```python
own_env: dict[str, str] = {}
env_var = server_config.env_var
if env_var:
    for key in credential_lookup_keys(server_config):
        value = os.environ.get(key)
        if value:
            own_env[env_var] = value
            break
return sanitized_subprocess_env(own_env)
```

Anything else the server needs must already be present in the gateway's own `os.environ` to survive `sanitized_subprocess_env`.

## Concrete case: self-hosted Firecrawl

`firecrawl-mcp` reads `FIRECRAWL_API_URL` to target a self-hosted instance instead of the SaaS endpoint. The manifest entry only declares `FIRECRAWL_API_KEY`:

```yaml
  firecrawl:
    requires_api_key: true
    env_var: "FIRECRAWL_API_KEY"
```

So a self-hosted deployment cannot be described by the manifest at all. The only way to connect one is to start the whole gateway with the variable pre-set:

```bash
FIRECRAWL_API_URL="http://<host>:3002" FIRECRAWL_API_KEY="self-hosted-no-auth" \
  pmcp --transport http --host 127.0.0.1 --port 3344
```

This works, but it is process-global: the setting applies to the gateway rather than to the server entry, it is invisible to anyone reading the manifest, and it is lost on any restart that does not reproduce the same environment. A self-hosted instance with no auth also has to be given a dummy API key purely to satisfy `requires_api_key`.

## Suggested direction

An optional `extra_env` mapping on the server entry, merged into `own_env` before `sanitized_subprocess_env`:

```yaml
  firecrawl:
    env_var: "FIRECRAWL_API_KEY"
    extra_env:
      FIRECRAWL_API_URL: "https://api.firecrawl.dev"
```

Worth deciding as part of the design:

- Whether values may reference secret storage keys, or are restricted to non-secret literals such as URLs (non-secret-only keeps the existing secret-handling and redaction paths untouched).
- Whether a user/project overlay may override `extra_env` per host — this is what makes a self-hosted endpoint practical, since the URL differs per machine while the shipped default stays the SaaS one.
- Whether `requires_api_key` should be relaxable when a self-hosted endpoint needs no credential, to avoid dummy values.

## Impact

Low severity, since the environment workaround exists. It mainly affects self-hosted or non-default endpoints, which currently cannot be represented in the manifest and so cannot be shared, reviewed, or version-controlled the way every other server setting is.

ViperJuice:
Part 2 of #105 — implements `plans/detailed-overlay-server-env-patch-20260801-1530.md`. Closes #105.

## Problem

`extra_env` (#108) made a non-secret variable declarable, but an overlay could only replace a server **wholesale**. Pointing a shipped server at a self-hosted endpoint meant hand-copying its `command`, `args`, four-platform `install` block, and `env_var` just to change one URL.

That copy is frozen. When the shipped entry later changes — package rename, install fix, new `secret_key` namespace — the overlay keeps overriding with the stale copy and the operator gets **no signal**, because the existing override warning fires on every override and cannot distinguish intent from staleness.

## Change

A `server_env:` overlay section that patches `extra_env` on an already-existing server, merging per key:

```yaml
# ~/.pmcp/manifest.yaml
server_env:
  firecrawl:
    FIRECRAWL_API_URL: "http://firecrawl.internal:3002"
```

Everything else about the shipped entry keeps flowing through.

## Why not deep-merge `servers:`

Whole-entry replace was an explicit prior decision (`plans/detailed-private-manifest-overlay-20260629-0841.md:23` — *"whole-entry replace, not deep-merge — simplest, predictable"*). It is **not** reversed here. Deep-merging would silently change behavior for anyone already using an overlay, and leaves "how do you *unset* an inherited field?" unanswerable. That plan file is amended in place to point at the new one, so the next reader of the decision finds the exception.

Interpolation (`${VAR}`) was rejected — it still needs the variable in the gateway env, the exact process-global workaround #105 exists to remove. An opt-in `merge: true` marker was deferred, not dismissed; this change doesn't foreclose it.

## Guardrails

- A patch naming an **unknown server warns and is skipped** — it can never conjure a server the manifest does not define.
- Patches apply **after** the same source's whole-entry replaces, so an overlay can replace an entry then refine it.
- Parsing reuses `_parse_extra_env`, so coercion and fail-soft behavior match `extra_env` exactly.

## One planned change deliberately dropped

The plan listed an optional explicit `FIRECRAWL_API_URL` default on the shipped firecrawl entry, marked *verify-or-drop*. **It fails verification.** `firecrawl-mcp` treats the variable as a presence sentinel for *"self-hosted, auth not required"*, not merely a URL:

| Location | Behavior when `FIRECRAWL_API_URL` is set |
|---|---|
| `dist/index.js:97` | Skips the `"Unauthorized: API key is required"` guard on monitor requests |
| `dist/index.js:1042` | Suppresses the keyless-mode notice |
| `dist/index.js:1047` | Bypasses the fatal missing-credential exit under HTTP streaming |

Shipping that default would have silently disabled firecrawl-mcp's own auth guards for every hosted user. Dropped; the mechanism works without it.

## Test plan

- [x] Full suite: **2267 passed**, 3 skipped (was 2259 — 8 new)
- [x] `tests/test_manifest_overlay.py`: 16 passed — **all 8 pre-existing tests pass unmodified**, demonstrating `servers:` semantics are unchanged
- [x] `ruff check` / `ruff format --check` / `mypy` clean
- [x] Edge cases verified against a real overlay file: empty-string values pass through (not dropped as falsy), falsy YAML scalars coerce (`0` → `"0"`, `false` → `"false"`), and a patch merges into a pre-existing `extra_env` without clobbering unrelated keys
- [x] New cases cover: shipped entry untouched; per-key merge; unknown server warns+skips; malformed `server_env` (non-mapping, and scalar value) skipped; cross-source precedence; replace-then-patch ordering; explicit `manifest_path` skips `server_env`

---
<a href="https://app.blacksmith.sh/Consiliency/codesmith/pmcp/pr/109"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a>


ViperJuice:
## Pre-merge advisor board — 1 blocking defect found and fixed

Board: `code-review`, 4 seats / 4 distinct vendors. **3 responded, 1 unavailable** (`claude:claude-fable-5:max:correctness` — `tui_adapter_required`, an environment limitation, not a review).

| Seat | Lens | Verdict |
|---|---|---|
| `grok-4.5:max` | adversarial | **DISAGREE** |
| `codex:gpt-5.6-sol:max` | red-team | **DISAGREE** |
| `gemini-3.6-flash:high` | alternative-approach | AGREE |

Both dissents landed on the **same defect, found independently**, and codex produced a direct exact-head reproduction. That is real signal, so it was treated as blocking.

### The defect

`extra_env` / `server_env` never reached the primary spawn path.

`_manifest_server_to_config` (`src/pmcp/config/loader.py`) built the runtime env from the resolved credential alone and never consulted `server.extra_env` — the string `extra_env` appeared **nowhere** in `config/loader.py`. Install-and-run builds its own child env via `build_install_child_env` and *did* carry it, which is exactly what made this hard to see:

1. First `gateway.provision` → install-and-run + adopt → URL present → self-hosted works.
2. Restart / refresh / lazy reconnect / `connect_server` → resolves through `_manifest_server_to_config` → spawns **without** the URL → **silent fallback to the vendor default**.

Codex's reproduction before the fix:

```text
install_has_url:     http://self-hosted.internal:3002
startup_runtime_env: {FIRECRAWL_API_KEY: ...}      # URL gone
```

This defeated the acceptance criterion of #105 itself. The 2267-test suite missed it because the overlay tests assert `ServerConfig.extra_env` after load and never the resolved `LocalMcpServerConfig.env`.

### The fix

Same class as #95 (config-source × injection), so I enumerated **every** site feeding `sanitized_subprocess_env` rather than patching only the reported one. Both are now fixed:

- **`_manifest_server_to_config`** — seed `env` from `extra_env`, then the credential on top so a resolved secret still wins a collision. `env or None` preserves the historical None-when-empty shape.
- **`_merge_manifest_defaults`** — a configured `.mcp.json` entry duplicating a manifest server inherits `extra_env` too, but an explicitly configured value wins, matching the override principle already used for credentials.

Same reproduction after the fix:

```text
manifest extra_env : {'FIRECRAWL_API_URL': 'http://self-hosted.internal:3002'}
install_has_url    : http://self-hosted.internal:3002
runtime_env        : {'FIRECRAWL_API_URL': 'http://self-hosted.internal:3002', 'FIRECRAWL_API_KEY': 'fc-key'}
```

### Non-blocking nits — all three addressed

- `_parse_extra_env` now takes a `field_label`, so an overlay patch warns about `server_env` rather than the `extra_env` shape it reuses internally.
- Added the untested cases the board named: empty-string patch value passes through (not dropped as falsy), and `$PMCP_MANIFEST_PATH` outranks the user overlay for `server_env`.

### Where the board agreed

All three responding legs independently confirmed **closed**: whole-entry-replace semantics for `servers:` are preserved (no deep-merge leak), source and replace-then-patch ordering is sound, credential shadowing / cross-server secret bleed / policy-excluded-server resurrection are all prevented, and fail-soft parity matches the existing overlay conventions.

### Verification after the fix

- Full suite: **2276 passed**, 3 skipped (was 2267 — 9 new)
- `ruff check` / `ruff format --check` / `mypy` clean

PR #115, opened 2026-08-05
chore(plan): stage-1 plan for the mcp 2.x / spec 2026-07-28 port

ViperJuice:
Planning artifact only — no code changes. Adds `plans/detailed-mcp-2x-spec-2026-07-28-stage1-20260805-1740.md` and its manifest entry.

## Why this exists

**mcp 2.0 is the only implementation of the current stable MCP spec (`2026-07-28`).** So the `mcp<2.0.0` cap from #111 — correct triage for a dead-on-arrival install — is now what pins pmcp to the previous protocol revision.

| | LATEST_PROTOCOL_VERSION | subscriptions |
|---|---|---|
| mcp 1.29.0 (current pin) | `2025-11-25` | ✗ |
| mcp 2.0.0 | **`2026-07-28`** | ✓ |

## Split into three stages

A full move has five distinct concerns, well past the bounded-plan threshold. Bundling them entangles the failure modes, and two of them change *observable client behaviour* while the rest are internal.

- **Stage 1 (this plan)** — runtime parity. Raise the cap, re-register handlers, extend the downstream version ladder. The gateway runs on mcp 2.x and behaves exactly as today.
- **Stage 2** — `server/discover` (new mandatory RPC). Additive.
- **Stage 3** — `subscriptions/listen` replacing the retired HTTP GET endpoint. Changes what clients see.

Stage 2/3 don't begin until Stage 1 has run for a normal working day.

## What the research turned up

Everything was verified by installing the libraries and running the code:

- **The import surface is nearly intact** — 6 of 7 `mcp` imports survive; only `streamablehttp_client` → `streamable_http_client`.
- **The real break is invisible to imports.** With that aliased, every module imports and the gateway then dies at boot: `'Server' object has no attribute 'list_tools'`. All six registrations are in one file.
- **`MCPServer` is a trap.** It has matching method names, so it looks like the migration target — but `StreamableHTTPSessionManager` takes the *lowlevel* `Server`, and `MCPServer.run()` owns its own transport, which would displace pmcp's Starlette app, `/health`, `/metrics`, and auth middleware. Stay lowlevel.
- **The hard part is already half-built.** `client/manager.py` already negotiates downstream with a fallback ladder (`PREFERRED_PROTOCOL_VERSION` `:187`, accepted list around `:190`, `_is_protocol_version_initialize_error` `:390`, retry at `2024-11-05` around `:1876`). Bridging 108 mostly-1.x servers is an extension of that, not a new subsystem.

## The verification lesson from #111, applied

The plan states explicitly that **the unit suite is not sufficient** — 2276 tests passed while the gateway could not boot. Stage 1's acceptance requires two runtime checks: the gateway actually listens, and a downstream 1.x server still serves a tool call through the 2.x gateway.

The `install-smoke` job from #111 must pass **unmodified**; if the port breaks it, the port is wrong, not the job.

## pangram-mcp ships separately

Its port is genuinely ~3 lines (`FastMCP` → `MCPServer`; `ToolAnnotations` still in `mcp.types`; `.tool()` still takes `annotations`). Deliberately not coupled — gating a three-line fix behind a gateway rewrite helps nobody.

---
<a href="https://app.blacksmith.sh/Consiliency/codesmith/pmcp/pr/115"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a>

### DIFF
--- a/plans/detailed-mcp-2x-spec-2026-07-28-stage1-20260805-1740.md
+++ b/plans/detailed-mcp-2x-spec-2026-07-28-stage1-20260805-1740.md
@@ -1,0 +1,116 @@
+# Detailed plan: port pmcp onto mcp 2.x (spec 2026-07-28) — Stage 1, runtime parity
+
+> **Plan Mode was NOT active when this was produced.** Planning artifact only — no implementation has begun.
+
+## Task
+
+Raise pmcp's `mcp<2.0.0` cap and get the gateway running on **mcp 2.x**, which is the only implementation of the current stable MCP spec (`2026-07-28`). mcp 1.29.0 tops out at `2025-11-25` and has no subscriptions support, so the cap added in #111 — correct as triage for a dead-on-arrival install — is now what pins pmcp to the previous protocol revision.
+
+**Stage 1 scope: run correctly on the mcp 2.x library while continuing to serve today's semantics.** New-spec *features* (`server/discover`, `subscriptions/listen`, retiring the HTTP GET endpoint) are explicitly out of scope — see "Why this is staged".
+
+## Research summary
+
+All facts below were verified in-session by installing the libraries and running the code, not by reading changelogs.
+
+- **mcp 2.0.0 is the 2026-07-28 implementation.** `LATEST_PROTOCOL_VERSION = 2026-07-28`; 1.29.0 reports `2025-11-25`. 2.0 adds `PROTOCOL_VERSION_META_KEY`, `UNSUPPORTED_PROTOCOL_VERSION`, and `UnsupportedProtocolVersionErrorData` — the primitives for per-request negotiation.
+- **The import surface is nearly intact.** Of pmcp's seven `mcp` imports, only `mcp.client.streamable_http.streamablehttp_client` is gone (renamed `streamable_http_client`). `sse_client`, `Server`, `StreamableHTTPSessionManager`, `SessionMessage`, and `Tool` all survive.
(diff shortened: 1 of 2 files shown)


ViperJuice:
Closing as **superseded and partly incorrect** — superseded by `specs/phase-plans-v11.md` (#116, merged as `df6efe0`), whose Phase 2 absorbs and corrects this plan.

**Do not implement from this document.** Its central instruction is wrong:

> `PREFERRED_PROTOCOL_VERSION` — raise to `"2026-07-28"` … **This is the single most important line in the plan**

Three advisor-board rounds on the roadmap established that `2026-07-28` **cannot be requested through `initialize`**. `mcp_types.version` separates the eras:

- `HANDSHAKE_PROTOCOL_VERSIONS = (2024-11-05, 2025-03-26, 2025-06-18, 2025-11-25)` — reached via `initialize`
- `MODERN_PROTOCOL_VERSIONS = (2026-07-28,)` — reached via per-request `_meta` + `server/discover`

`PREFERRED_PROTOCOL_VERSION` governs the **downstream handshake ladder**, whose ceiling of `2025-11-25` is correct. Following this plan would have set an unreachable version on the wrong side of the gateway.

Two further errors in this document, both corrected in v11 P2:

- **"Handler bodies do not change; only registration moves."** False. 2.x invokes `handler(ctx, typed_params)` and validates a typed result; the removed decorators supplied argument adaptation, result wrapping, and tool-input-schema validation. Explicit adapters are required.
- **"Rename the `streamablehttp_client` import and its call site."** Understated. `headers`, `timeout`, `sse_read_timeout`, and `httpx_client_factory` are gone in favour of a caller-built `httpx2` client that 2.x deliberately does not close — so the naive rename `TypeError`s on every authenticated remote downstream, and the fix leaks a client per reconnect without exit-stack ownership.

The correct version of this work is v11 **P2**, planned via `/claude-plan-phase P2` after P1.

PR #122, opened 2026-08-06
feat(P1): release-guard hardening — min-version-smoke, schedule, drift test, honest mcp floor

ViperJuice:
Executes **v11 Phase 1 (P1)** per `plans/phase-plan-v11-P1.md`. First executed phase of the roadmap.

## What ships

Four release guards, ported from `Consiliency/pangram-mcp`:

| Guard | Status |
|---|---|
| `install-smoke` | **byte-for-byte unchanged** — verified programmatically |
| `min-version-smoke` | new |
| `schedule:` + `workflow_dispatch:` | new |
| `__version__` / metadata drift test | new |

Plus a correction to a **shipped dependency bound**.

## pmcp has been declaring a false `mcp` floor

`pyproject.toml` said `mcp>=1.0.0`. Established by installing each candidate, not by reading source:

| mcp | `import pmcp.client.manager` |
|---|---|
| 1.0.0 | ❌ `No module named 'mcp.client.streamable_http'` |
| 1.7.1 | ❌ same |
| 1.8.0 | ✅ |

`client/manager.py:22` imports from that module. Now `mcp>=1.8.0,<2.0.0`.

The roadmap listed "changing any dependency bound" as a P1 non-goal, but held literally that makes the phase unmergeable — `min-version-smoke` at `>=1.0.0` lands red on `main`, and cross-cutting principle 2 forbids weakening the job to make it pass. **The roadmap amendment ships in this same commit**, per the plan: `pyproject.toml`, `uv.lock`, and `specs/phase-plans-v11.md` are not separable.

This is the third instance of this bug class this session — pmcp's missing upper bound shipped broken twice (#111), and pangram-mcp's lower bound was wrong by twelve minor versions (0.1.3).

## `min-version-smoke` proves a real tool call, not an import

Imports are not acceptance for a gateway — 2276 tests once passed while it could not boot. Neither is `/health`, which returns `"ok": True` as a hardcoded literal (`transport/http.py:435-436`).

So the job boots at the declared floor and drives a **client → gateway → stdio downstream → back** round trip through `.github/probe/`.

Its isolation is the part that matters, because `_kill_orphan_processes` (`server.py:683-700`) SIGKILLs every `/proc` match among resolved configs:

- `HOME` + `XDG_CONFIG_HOME` redirected — load-bearing, since `--config` alone does not bound resolution (`load_manifest()` starts from the package-shipped manifest)
- `--policy` allowlisting only the probe — this is what actually bounds the kill set
- `--lock-dir`, `--config`, `--project`, port 3399, **never 3344**
- asserts `Gateway initialized: N/1 servers online`, with a comment recording that `skipped=106, policy_denied=106` is what a *correct* run looks like so the tripwire is not re-inverted
- `$!` captured, `trap`-cleaned

## Verification

```
install-smoke byte-identical to main:  True   (job block extracted and compared)
pytest                                 2277 passed, 3 skipped   (baseline 2276)
ruff check / format                    clean
mypy src/                              clean, 41 files
actionlint                             clean
FLOOR reaches the shell via env:       yes, never inlined into run:
live gateway on :3344                  healthy throughout
```

Reviewed by three advisor-board rounds at the plan stage before any code was written.

---
<a href="https://app.blacksmith.sh/Consiliency/codesmith/pmcp/pr/122"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a>

### DIFF
--- a/.claude/docs-catalog.json
+++ b/.claude/docs-catalog.json
@@ -1,1 +1,322 @@
-[]
+{
+  "version": 1,
+  "last_updated": "2026-08-06T03:37:43+00:00",
+  "files": {
+    "README.md": {
+      "purpose": "root project README",
+      "touched_by_phases": []
+    },
+    "CHANGELOG.md": {
+      "purpose": "release changelog",
+      "touched_by_phases": [
+        "P1"
+      ]
+    },
+    "CONTRIBUTING.md": {
+      "purpose": "contribution guidelines",
+      "touched_by_phases": [
+        "P1"
+      ]
+    },
+    "SECURITY.md": {
+      "purpose": "security policy",
+      "touched_by_phases": []
+    },
+    ".claude/plans/ticklish-inventing-stearns.md": {
+      "purpose": "markdown document",
+      "touched_by_phases": []
+    },
+    "specs/README.md": {
+      "purpose": "root project README",
+      "touched_by_phases": []
+    },
+    "specs/active/pmcp-gateway-orchestration-plan.md": {
+      "purpose": "markdown document",
+      "touched_by_phases": []
+    },
+    "specs/phase-plans-v1.md": {
+      "purpose": "multi-phase roadmap spec",
+      "touched_by_phases": []
+    },
+    "specs/phase-plans-v10.md": {
+      "purpose": "multi-phase roadmap spec",
+      "touched_by_phases": []
+    },
+    "specs/phase-plans-v11.md": {
+      "purpose": "multi-phase roadmap spec",
+      "touched_by_phases": [
+        "P1"
+      ]
+    },
+    "specs/phase-plans-v2.md": {
+      "purpose": "multi-phase roadmap spec",
+      "touched_by_phases": []
+    },
+    "specs/phase-plans-v3.md": {
+      "purpose": "multi-phase roadmap spec",
+      "touched_by_phases": []
+    },
+    "specs/phase-plans-v4.md": {
+      "purpose": "multi-phase roadmap spec",
+      "touched_by_phases": []
+    },
+    "specs/phase-plans-v5.md": {
+      "purpose": "multi-phase roadmap spec",
+      "touched_by_phases": []
+    },
+    "specs/phase-plans-v6.md": {
+      "purpose": "multi-phase roadmap spec",
+      "touched_by_phases": []
+    },
+    "specs/phase-plans-v7.md": {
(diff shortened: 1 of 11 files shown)


ViperJuice:
## Pre-merge board review

Board: `code-review`, 4 seats. **2 usable legs** (`claude-fable-5` UNAVAILABLE `tui_adapter_required`, gemini EMPTY) — below the independence floor, so this is a two-reviewer split rather than a quorate verdict.

| Leg | Verdict |
|---|---|
| `grok-4.5:max` adversarial | all six review questions **sound** |
| `codex:gpt-5.6-sol:max` red-team | **DISAGREE** — one blocking defect |

### The defect, confirmed and fixed

`.github/probe/p1_probe_client.py` checked only `connected.isError`. But `gateway.connect_server` does not signal failure that way — it returns `LifecycleServerOutput` serialized as an ordinary JSON text block, so a genuine failure arrives as `{"ok": false, "message": "..."}` with **`isError` false**.

Verified against real behaviour observed in this session:

```json
{"ok": false, "server": "firecrawl", "new_status": "lazy",
 "message": "Server 'firecrawl' requires authentication. Set one of: FIRECRAWL_API_KEY"}
```

delivered as normal content, not an error.

The compounding part: a failed explicit connect leaves the fixture registered as **lazy**, so the following `gateway.invoke` reconnects on demand and returns the sentinel anyway. **The job would go green while `gateway.connect_server` was broken** — the exact "passes while something real is broken" failure this phase exists to prevent, and a contradiction of IF-0-P1-1's claimed *connect-plus-invoke* proof.

Fixed in `f4f386e`: a `_check()` helper parses the JSON body and requires `ok is True`, applied to **both** the connect and the invoke. The `isError` check is retained — it catches transport-level failures that never produce a JSON body at all. The rationale is recorded in the docstring so the weaker check is not reintroduced.

### Independently verified after the fix

```
install-smoke byte-identical to main:  True   (job block extracted and compared)
_check applied to both gateway calls:  yes
probe client parses:                   ast.parse clean
CI on f4f386e:                         9/9 pass, including min-version-smoke
live gateway on :3344:                 healthy throughout
```

PR #154, opened 2026-08-16
fix: unify update_server config resolution across notice paths and the probe window

### LINKED ISSUE #150: update_server and the stale-notice paths resolve server config differently
Found during board review of #147 (round 3). Not a blocker for that PR, but a real inconsistency it leaves behind.

## What

`gateway.update_server` now resolves its target through `_resolve_lifecycle_config`, which gives `.mcp.json` precedence over the manifest — the same precedence `gateway.restart_server` uses. The **notice** paths were not migrated and still use the manifest-only resolver:

- `handlers.py:1324` — the background stale sweep
- `handlers.py:3596` — `_get_update_warning`, via `_get_server_config_for_update` (`handlers.py:3586`)

## Why it matters

After a successful update, `update_server` seeds `_stale_check_cache` with the override package's `(current, latest)` pair. That entry is honoured for the TTL (6h). Once it expires, the notice paths recompute — but they resolve the **manifest** package, and compare its upstream latest against the version recorded from the **override** package.

When the override merely pins the same package, the two agree and nothing goes wrong. The problem case is an override that changes package identity: the comparison is then between unrelated packages, which can resurface a false "update available" notice or hide a real one.

## Fix sketch

Give the notice paths the same effective-config resolution `update_server` now uses, so all four sites (`_get_update_warning`, `catalog_search`'s `stale_updates`, the background sweep, and `update_server`) agree on which package a server actually runs. Care needed: `_get_server_config_for_update` returns a manifest `ServerConfig` while `_resolve_lifecycle_config` returns a `ResolvedServerConfig`, so it isn't a drop-in swap.

## Blast radius

Only servers whose `.mcp.json` entry names a different package than the manifest entry of the same name. Unpinned and same-package-pinned overrides are unaffected.

### LINKED ISSUE #151: update_server resolves config twice across the probe window (TOCTOU)
Found during board review of #147 (round 3). Narrow race, filed rather than fixed.

## What

`gateway.update_server` resolves the server's effective config via `_resolve_lifecycle_config` (around `handlers.py:5033`), then runs the update probe — which has a **60 second** timeout — and only afterwards calls `self.restart_server(...)`, which independently resolves the config *again* (`handlers.py:2888`).

Both calls go through the same function with the same precedence, so under a stable configuration they cannot disagree. But they are two separate reads separated by a potentially long await.

## Why it matters

If `.mcp.json` is edited during the probe window, `update_server` can probe and install package A, restart onto package B, and then persist A's version and A-derived bookkeeping for B. That is the same silent-misreport class #147 exists to eliminate, reachable through a TOCTOU window rather than through resolver divergence.

The PR's own code comments claim the config is "resolved once", which is true of the *function* but not of the *read*. Worth correcting either the comment or the behaviour.

## Fix sketch

Thread the already-resolved `ResolvedServerConfig` through to the restart instead of re-resolving — e.g. call `self._client_manager.restart_server(resolved_config.config, force=...)` directly. The tradeoff: the public `restart_server` handler also performs policy checks, credential-gap detection, and remote-header validation, so bypassing it would mean either duplicating those or accepting their loss. An alternative is to re-resolve and compare against the first result, refusing if it changed.

## Blast radius

Requires a config edit landing inside the probe window. Unlikely in normal operation, plausible during automated config rollouts.

ViperJuice:
Closes #150. Closes #151. Both are residuals filed during the PR #147 board review.

## #150 — stale notices could compare two different packages

`gateway.update_server` resolves its target with `.mcp.json` taking precedence over the manifest; the two notice paths (`_get_update_warning` and the background stale sweep) still resolved manifest-only. For a server configured in both places under one name they version-checked a package that was not the one running.

Resolving the lookup turned out to be **half the fix**. The *current* side of the comparison comes from the descriptions cache, generated from whichever package that cache last described — so pointing only the *latest* lookup at the effective package still compares unrelated packages. `GeneratedServerDescriptions` records the package it describes, so the mismatch is detectable: both paths now stay silent until `pmcp refresh` re-describes the server, rather than emitting a comparison they cannot justify.

Two deliberate design points:

- **Not** reusing `_resolve_lifecycle_config`, despite the issue's own fix sketch suggesting it. That function returns `LifecycleServerOutput` *failure* objects for policy denial, missing headers, and credential gaps — right for a lifecycle action, wrong for a best-effort notice that has no channel for a failure output and must fail silent.
- An explicit `is_server_allowed` check, because `_load_all_configured_servers` returns entries *before* policy filtering (its own docstring says so). Without it a denied server leaks notices. It is a pure predicate, so the helper stays side-effect free. Credential gaps are deliberately not checked — they do not change which package is installed, so a notice about it is still truthful.

## #151 — a config edit during the probe could activate the wrong package

The probe can run for up to 60 seconds, and `gateway.refresh()` — which runs between the probe and the restart — reloads configuration from disk. The original plan was to pin the pre-probe config through to the restart; the board caught that this is **worse** than the race, since it would disconnect the edited server and reconnect the old package, and would bypass restart-time credential gating.

The update now re-resolves after the probe and requires the effective command/args to be unchanged. On drift it aborts with `ok: false` and records no version state. The confirmed config is handed to `restart_server` via a new internal keyword-only `resolved_config` parameter so the restart cannot resolve a third time. The public tool schema is unchanged.

## Planned and boarded first

Plan committed at `.consiliency/plans/detailed-update-server-resolution-residuals-20260816-1824.md`. The advisor panel (grok-4.6 / codex / gemini) reviewed it **before** implementation and found three blocking defects in the plan — the half-fix above, the #151 pin-through-the-window mistake, and an unsatisfiable policy-silence criterion. All three were verified against source and the plan was revised; both versions are in the git history.

## Verification

- `pytest -q` → **2543 passed, 3 skipped, 0 failed** (2536 baseline + 7 new tests)
- `ruff check .`, `ruff format --check src/ tests/`, `mypy src` — all clean
- **RED proof**: reverting only `src/pmcp/tools/handlers.py` (tests untouched) turns all 6 new behavioural tests red. The failures are the actual bugs, not assertion noise:
  - #150 → `assert "Update available for 'playwright': 0.1.0 -> 99.0.0. ..." is None` — a notice invented by comparing the cached package's version against a different package's latest
  - #151 → `assert True is False` on `UpdateServerOutput(ok=True, ... 'the new version is now active')` while the config had changed underneath

Tests assert observable outcomes (was a notice emitted, what did the cache record), not argument plumbing — per the board, a "package B was queried" assertion passes while the bug survives.

---
<a href="https://app.blacksmith.sh/Consiliency/codesmith/pmcp/pr/154"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789503189&installation_model_id=427051&pr_number=154&repository=Consiliency%2Fpmcp&return_to=https%3A%2F%2Fgithub.com%2FConsiliency%2Fpmcp%2Fpull%2F154&signature=7aa2dcb18e14df10384a0bd0daf198c4442a35fbe3ac9b40d9f8bb40157c9d7f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/.consiliency/plans/detailed-update-server-resolution-residuals-20260816-1824.md
+++ b/.consiliency/plans/detailed-update-server-resolution-residuals-20260816-1824.md
@@ -20,24 +20,39 @@ Both notice call sites consume only `.command` and `.args` (verified at `handler
 
 For #151, `restart_server` (`handlers.py:2908`) resolves and then calls `self._client_manager.restart_server(config, force=...)`. The `ClientManager` method takes an already-resolved config, so threading the caller's config through avoids the second resolution without duplicating the handler's policy/credential gating.
 
+## Board review — three blocking defects found, plan revised
+
+This plan was reviewed by the advisor panel (grok-4.6 AGREE, codex DISAGREE, gemini DISAGREE) before implementation. Two legs independently found the same policy-gating contradiction; codex found two further defects. **All three were verified against source and the plan below is the revised version.** Recorded here because the original design would have shipped an incomplete fix.
+
+**B1 — the notice fix was half a fix (codex).** Changing only the *latest* lookup to the configured package leaves *current* as `GeneratedServerDescriptions.version`, which was generated from the **manifest** package (`refresher.py`). So manifest package A at `10.0.0` vs configured package B at `2.0.0` still compares unrelated packages and can hide B's real update. `GeneratedServerDescriptions` carries a `package` field (`types.py:1365`) that records which package the entry describes — the original plan never consulted it. A test asserting "B was queried" would have passed while the bug survived.
+
+**B2 — the #151 fix could restart the wrong package (codex).** `update_server` calls `self.refresh(...)` *after* the probe (`handlers.py:5167`), which reloads config and applies an edit to B. Passing the stale probed A into the restart would then disconnect B and reconnect A — worse than the TOCTOU it fixes — and would bypass restart-time credential gating.
+
+**B3 — the policy-silence criterion was unsatisfiable (codex + gemini).** `_load_all_configured_servers` returns entries *before* policy filtering (its own docstring says so). A helper doing no gating would return command/args for a denied server, so `test_notice_paths_stay_silent_for_policy_denied_server` could not pass as specified, and denied servers would leak notices.
+
 ## Changes
 
 ### `src/pmcp/tools/handlers.py` (modify)
 
-- `_effective_server_config_for_notice` — **add** — new private helper returning `tuple[str, list[str]] | None` (command, args) for a server, applying the same `.mcp.json`-over-manifest precedence as `_load_all_configured_servers` and falling back to manifest → discovered exactly as `_get_server_config_for_update` does today. Returns `None` when nothing resolves, when the effective config is remote (no `command`), or when the command is empty. Deliberately performs **no** policy/credential/header gating — a notice is best-effort and must fail silent, not emit a lifecycle failure.
-- `_run_stale_index` (call site around `handlers.py:1350`) — **modify** — replace `self._get_server_config_for_update(server_name)` + `.command`/`.args` access with the new helper; keep the existing `continue`-on-`None` behavior.
-- `_get_update_warning` (call site around `handlers.py:3622`) — **modify** — same replacement; keep the existing `return None`-on-`None` behavior.
-- `_get_server_config_for_update` — **keep, unmodified** — still used by `update_server`'s own manifest-oriented lookups; not deleted, and its docstring gains one line noting notice paths use the effective-config helper instead.
-- `update_server` (around `handlers.py:5059`) — **modify** — after resolving `resolved_config`, pass it to the restart rather than letting `restart_server` re-resolve (see next item), closing the #151 TOCTOU window.
-- `restart_server` (`handlers.py:2908`) — **modify** — accept an optional pre-resolved config so a caller that already resolved can supply it. Signature stays `async def restart_server(self, input_data: dict[str, Any])` for the public tool contract; add a keyword-only `resolved_config: ResolvedServerConfig | None = None` parameter used in place of the internal `_resolve_lifecycle_config` call when supplied. When `None`, behavior is byte-identical to today.
+- `_effective_notice_target` — **add** — new private helper returning `tuple[str, list[str]] | None` (command, args). Applies `.mcp.json`-over-manifest precedence via `_load_all_configured_servers`, falling back to manifest → discovered as `_get_server_config_for_update` does today. Returns `None` when: nothing resolves; the effective config is remote (no `command`); the command is empty; **or the server is not `self._policy_manager.is_server_allowed(...)`** — a pure predicate (`policy.py:104`), so this is the side-effect-free eligibility check B3 requires, without importing `_resolve_lifecycle_config`'s failure-output machinery. Credential/header gaps are deliberately *not* checked: they do not change which package is installed, and a notice for a credential-gapped server is still truthful.
+- `_notice_package_matches_cache` — **add** — returns whether the effective config's detected package equals the `package` recorded on the descriptions-cache entry. Fixes **B1**: `_get_update_warning` and `_run_stale_index` must compare like with like. Uses the existing `detect_package_type` to derive the effective package name.
+- `_run_stale_index` (call site around `handlers.py:1350`) — **modify** — resolve through `_effective_notice_target`; **skip the server entirely when the package identity does not match** the cache entry, rather than writing a cross-package `(current, latest)` pair. Keeps the existing `continue`-on-`None`.
+- `_get_update_warning` (call site around `handlers.py:3622`) — **modify** — same resolution and the same identity guard; return `None` on mismatch (fail silent — a wrong notice is worse than none). Keeps the existing `return None`-on-`None`.
+- `_get_server_config_for_update` — **keep, unmodified** — still used by `update_server`'s manifest-oriented lookups; docstring gains one line pointing notice paths at the new helper.
+- `update_server` (around `handlers.py:5059`) — **modify** — fixes **B2**. After the probe and `refresh()`, **re-resolve** the config and compare it against the probed snapshot (command + args). If it changed, abort with `ok=False` and a message naming the drift, and mutate **no** cache or stale-check state. Only when unchanged, pass the (now confirmed-current) config to the restart. This closes #151's window by *detecting* the edit rather than by pinning a stale config through it.
+- `restart_server` (`handlers.py:2908`) — **modify** — add a keyword-only `resolved_config: ResolvedServerConfig | None = None`, used in place of the internal `_resolve_lifecycle_config` call when supplied. `update_server` passes the re-resolved-and-confirmed config. When `None`, behavior is byte-identical to today. Note per gemini: this is safe *because* `update_server` performs the same gating first; the parameter must stay internal and never reach the public tool schema.
 
 ### `tests/test_tools.py` (modify)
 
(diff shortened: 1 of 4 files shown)


ViperJuice:
## Board review: 2 AGREE / 1 DISAGREE — not merging; the dissent is right

Panel run on the finished implementation (grok-4.6 / codex / gemini, 3/3 delivered). grok and gemini returned AGREE; codex returned DISAGREE with three findings. I verified the first against source and **it holds**, so this PR stays open rather than merging on a 2-1 count.

### Confirmed: warm cache entries still leak cross-package notices via `catalog_search`

My fix makes the two notice paths *skip* on package-identity mismatch — but skipping is not invalidating. `catalog_search` reads `_stale_check_cache` directly:

```python
for sn, (_, current, latest) in self._stale_check_cache.items()
if current and latest and is_version_newer(current, latest)
and self._policy_manager.is_server_allowed(sn)
```

No TTL check, no package-identity check. So the sequence that still breaks: package A populates the cache → config switches to B → the sweep and `_get_update_warning` correctly go silent, but the **A-derived tuple stays in the cache** and `catalog_search` keeps attributing it to B indefinitely.

That is the same defect #150 describes, surviving in the one emission site I did not close. My new tests all start from an empty cache, so none of them could have caught it — the exact "test passes while the bug survives" shape the plan review warned about, reproduced by me one layer down.

The fix is to **invalidate** rather than skip: `pop` the entry on identity mismatch so no downstream reader can serve it, plus a test that pre-populates the cache with an A-derived tuple, switches the config to B, and asserts `catalog_search` emits nothing for that server.

### Also raised, and worth taking seriously

**`command + args` is not sufficient process identity** (#151). The repo's own `_refresh_config_unchanged` compares `cwd` and resolved `env` too. A change to `DOCKER_HOST`, or to package-manager registry/cache env, can leave command+args equal while the probe and the restart operate against different installations. My drift check should compare the same fields that function does.

**A second await still follows the recheck.** `get_package_version` runs after the config recheck, reopening a smaller window before the restart. Narrower than the original 60s probe, but the same class.

**The advertised recovery path may not exist.** I told users a mismatch resolves once `pmcp refresh` re-describes the server — but `refresh_all` describes *manifest* entries, not the effective `.mcp.json` override, so an override-named package may never produce a matching cache entry. If that holds, "stay silent until refresh" means "stay silent forever" for those servers, and either the recovery path or the CHANGELOG wording needs to change. I have not traced this one end to end.

### Status

Not merging. The identity-invalidation fix and the `cwd`/`env` comparison are both small; the recovery-path question needs tracing first because it may change what the fix should *do* on mismatch rather than just how it detects one.

Worth noting for the record: the plan for this change was boarded before implementation and the panel found three blocking defects then too. Four review rounds on PR #147, one on the plan, one here — every round has found something real in this subsystem.

ViperJuice:
## Round-3 board: unanimous DISAGREE (3/3) — my "self-heal" does not heal

All three legs blocked at head `85918b1`. I verified the two findings against source and reproduced the main one directly. **Both hold, and the second is the same failure you were told this rework fixed.**

### Blocking 1 — the recovery never establishes a baseline, so silence is still permanent

Re-pointing sets `package=B, version=""`. But `_get_update_warning` returns early on an empty version, *before* any lookup, and `_run_stale_index` skips fresh entries *before* resolving the effective target. So nothing ever observes a version for B.

Reproduced at exact head, three consecutive passes:

```
pass1: None    entry -> @new/pkg ''
pass2: None    entry -> @new/pkg ''
pass3: None    -> permanently silent
```

Worse, codex found the erasure path: a successful override update writes B's version, and the *next* warning re-points and wipes it — `('@playwright/mcp','9.9.9')` immediately after success, then `('different-playwright-fork','')`. Real updates to B stay hidden indefinitely.

This is precisely the outcome the rework was supposed to eliminate. I changed the mechanism (re-point instead of stay silent) and reproduced the behaviour (never notifies again). **My test asserted only that `entry.package` changed — never that a notice actually resumes**, so it passed while the bug survived. That is the identical test-shape flaw I flagged in round 1 and then repeated.

### Blocking 2 — warm tuples still leak through `catalog_search`

`_run_stale_index` skips fresh entries before resolving, and `catalog_search` still validates neither TTL nor package identity. Eviction only happens on paths that actually reach the identity check. Codex reproduced both a local A→B override and a local→remote override retaining an A-derived tuple and emitting the false A notice. My warm-cache test only drove `_get_update_warning`, and my remote test started from an empty cache — so both passed without covering the acceptance criterion.

### Assessment

Three rounds, three sets of real defects, and my last two fixes each shipped a hole of the same class the round before had just taught me to look for. The pattern is consistent: the descriptions cache and `_stale_check_cache` are being asked to carry a package identity they were never designed to key on, and every fix so far has patched one reader while another kept serving the stale tuple.

I do not think a fourth patch at this layer is the right move, so I am not attempting one. Not merging; taking the design question back to the maintainer.

What a real fix likely needs, for whoever picks this up:
- key `_stale_check_cache` on `(server, package)` rather than `server`, so a cross-package tuple is structurally unreachable instead of needing eviction at each reader; and
- give `catalog_search` the same freshness/identity contract as the other readers, rather than letting it read raw tuples.

Both are larger than this PR's scope. The plan doc, all three board rounds, and the reproductions are on the branch and in this thread.

ViperJuice:
Closing this PR unmerged. Replaced by a structural plan, committed as `.consiliency/plans/detailed-stale-notice-package-identity-20260817-0300.md` (`ef81638`).

Three board rounds each found real defects here, and my last two fixes each shipped a hole of the same class the previous round had just exposed:

- **Round 1** — skipping on package mismatch without evicting left `catalog_search` serving the poisoned tuple. My tests all started from an empty cache, so none could catch it.
- **Round 2** — the "self-heal" re-point set `version=""`, and every reader returns early on an empty version. I reproduced it live: three consecutive passes, three `None` — permanently silent, which is precisely the outcome the rework was chosen to avoid. A successful update's version baseline was then erased by the next notice check.
- **Round 3** — unanimous DISAGREE, plus a local→remote override leak I had not covered.

The pattern is the diagnosis. `_stale_check_cache` is keyed by **server name alone** while holding versions that belong to a **specific package**, and its three readers apply three different contracts:

| Reader | TTL | Identity |
|---|---|---|
| `_run_stale_index` | yes — but checked *before* resolving the target | patched |
| `_get_update_warning` | yes | patched |
| `catalog_search` | **no** | **no** |

So every fix patched one reader while another kept serving the stale tuple, and the sweep's TTL-before-resolve ordering meant a poisoned-but-warm entry was never re-examined for its full 6h life.

The replacement plan re-keys the cache on `(server_name, package)`, which makes a cross-package entry **unrepresentable** rather than something each reader must remember to detect, gives `catalog_search` the same TTL contract as the other two, and reorders the sweep to resolve before checking freshness. It also drops the `_repoint_...` helper entirely — that was the mechanism that erased good version baselines.

What carries forward unchanged, having been accepted in every round: the `.mcp.json`-over-manifest precedence with its `is_server_allowed` gate, and the whole #151 drift fix (recheck after the probe *and* after the version lookup, compared via `_refresh_config_unchanged` so `cwd` and effective `env` count, not just argv).

#150 and #151 both remain open. The plan, all three review transcripts, and the live reproduction are on record here and on the `fix/150-151-resolution` branch, which I am leaving in place rather than deleting so the failed approaches stay inspectable.

PR #155, opened 2026-08-17
fix: version comparison fails closed instead of fabricating update notices

ViperJuice:
Prerequisite for #150. Independently useful and independently testable, so it lands on its own.

## The bug

`is_version_newer` extracted every digit from both strings and compared the results, with no way to signal "this is not a version". Returning `True` makes the gateway tell an operator their server is out of date, so an unreadable version could only ever produce a false claim. Reproduced on `main`:

```
is_version_newer('nightly',           '2.0.0') -> True
is_version_newer('release-channel-a', '2.0.0') -> True
is_version_newer('build-1',           '2.0.0') -> True
is_version_newer('',                  '2.0.0') -> True
is_version_newer('1.0.0',      'sha256:abc123') -> True
```

Each of those emits an "update available" notice that is not true.

**The empty-string case is reachable in ordinary use**, not a contrived input: `MCPServer.__init__` in the current MCP SDK defaults `version` to `""`, and FastMCP 1.x reported the *SDK* version rather than the package's. Both were already documented in this repo (`plans/phase-plan-v11-PG.md:44`).

## The fix

Recognise a release number specifically — `1.2.3`, `1.0`, `2025.12.18`, a `v` prefix, and pre-release/build suffixes like `-rc1` or `+build.5` — and report **no update** when either side is not one.

Two subtleties worth calling out:

- **"Contains a digit" is not enough.** `build-1` parses to `(1,)` under a digits-anywhere scan and compares as older than every real release. The predicate has to anchor on a *leading* numeric release component, or this specific fabrication survives. My first attempt at this fix got it wrong and the behaviour matrix caught it.
- **Docker would otherwise go silent.** `get_docker_version` returns a `sha256:` digest, which is version-shaped by no definition. Digests are now compared for **inequality** — a different digest genuinely is a new image, while ordering hex is meaningless — so the docker lane keeps working instead of being silenced by the stricter rule.

Existing documented behaviour is preserved, including `1.0.0-rc1` -> `1.0.0-rc2` = `True` and the `test_non_numeric_versions` contract that `alpha`/`beta` are mutually not-newer.

## Verification

- `pytest -q` -> **2541 passed, 3 skipped, 0 failed** (2536 baseline + 5 new tests)
- All 69 pre-existing `test_version_checker.py` tests still pass, unmodified
- **RED proof**: reverting only `src/pmcp/manifest/version_checker.py` fails 3 of the new tests, including `is_version_newer('1.0.0', 'sha256:abc123') -> True`

## Why this is separate from #150

#150 needs a version source whose identity is bound to the package being compared — five attempts have failed at that, and the analysis is recorded across two plan artifacts and a closed PR (#154). But *whatever* source is chosen, it will fabricate notices while the comparator cannot fail. This makes the rest sound.


ViperJuice:
## Board: 3/3 DISAGREE — not merging. All three findings reproduce.

I verified each against source and reproduced them at PR head. **Every one holds, and the second is the same class of mistake I opened this PR to fix.**

### 1. The fail-closed gate still accepts malformed versions

`_VERSION_RE`'s suffix group `(?:[-+.].*)?` swallows everything after the first `-`, `+` or `.`, and `_parse_version_parts` then extracts every digit:

```
is_version_newer('2026-08-17-nightly', '3000.0.0') -> True    # parses as (2026, 8, 17)
is_version_newer('1.not-a-version',    '2.0.0')    -> True
```

`2026-08-17-nightly` is the string my own test declares "unreadable". It passed **only because I compared it against `2.0.0`**, where `(2026,8,17) > (2,0,0)` happens to give `False`. Change the operand to `3000.0.0` and the fabrication reappears. That is the fourth — now fifth — time a test of mine has passed while its bug survived.

### 2. My docker handling silences docker entirely

This is the one that stings. `get_docker_version` **strips the `sha256:` prefix** and returns a bare 12-hex string (`version_checker.py:248-252`, pinned by the existing test at `test_version_checker.py:697`). My `_is_digest` only matches values that *start with* `sha256:`, so it never matches the real producer:

```
_is_digest('abcdef123456')                        -> False
is_version_newer('abcdef123456','fedcba654321')   -> False   # real digest change, SILENCED
```

I wrote that guard specifically to keep the docker lane working, tested it against `sha256:`-prefixed values **I invented**, and never checked what the producer actually returns. The result is exactly the failure I claimed to prevent — and the PR body asserts the opposite in as many words.

### 3. Accepted semver suffixes are ordered wrongly

Digit extraction ignores semver precedence:

```
is_version_newer('1.0.0-rc1', '1.0.0')          -> False   # hides a real stable release
is_version_newer('1.0.0',     '1.0.0-rc1')      -> True    # fabricates: a prerelease is NOT newer
is_version_newer('1.0.0+build.4','1.0.0+build.5') -> True  # build metadata is not precedence
```

This is pre-existing behaviour that my change preserved (the `-rc1`/`-rc2` case is pinned by an existing test), but the PR claims prerelease support and a fail-closed contract, and it delivers neither.

### Assessment

The right fix is almost certainly to stop hand-rolling this: parse with a real implementation (`packaging.version` for PEP 440, plus explicit semver precedence) and treat a parse failure as "no update". Hand-rolled digit extraction has now produced a fabricated-notice bug in every form it has taken.

Also needed regardless: `_is_digest` must match what `get_docker_version` actually emits, and the docker test must drive the **producer**, not a synthetic literal.

Not merging. I am not attempting a rewrite on top of five failed attempts in this subsystem without a maintainer decision on the dependency question, since `packaging` is a new runtime dependency.

ViperJuice:
## Round 4: 1 AGREE / 1 DISAGREE / 1 no-verdict — three findings, all real, all fixed

Verified each against source before fixing. **One was a regression this PR introduced**, and it is the most instructive of the three.

### 1. My fail-closed change silently inverted a negated caller — REGRESSION

`refresher.py:219` does `not is_version_newer(existing_cache.version, version, pkg_type)` and treats the result as "up to date". Those are not the same thing once the comparator fails closed.

Measured on `main` vs this branch:

```
main:        is_version_newer('unknown','2.0.0') -> True   =>  not(...) = False  => cache regenerates  (correct)
this branch: is_version_newer('unknown','2.0.0') -> False  =>  not(...) = True   => stale cache kept   (broken)
```

And `refresher.py:233` persists the literal `"unknown"` after a failed version lookup — so that cache would be pinned **forever**, never regenerating even once the lookup started succeeding.

This is the hazard of making a predicate stricter: every negated call site silently flips meaning. Fixed by adding `is_version_orderable()` and gating the call site on it, so "not newer" is only trusted as "up to date" when the cached value is actually orderable.

### 2. My hex-letter rule rejected valid all-numeric digests

Last round I required a hex letter so a numeric calendar version (`202612180000`) would not be misread as a digest. But `get_docker_version` truncates SHA-256 to 12 hex chars, and that can be all digits — so `987654321098` → `123456789012`, two genuinely different images, reported **no update**.

Shape alone cannot separate the two cases. `package_type` now decides: with `docker`, any hex shape is a digest; without it, a hex letter is still required so numeric versions keep being ordered. Both directions are tested.

### 3. `_SEMVER_RE` ordered strings SemVer forbids

It accepted leading zeros in core and prerelease identifiers (spec rules 2 and 9) and empty identifiers (rule 10), then ordered them — so `1.0.0-01` fabricated an upgrade to `1.0.0`. These cannot be published to npm. The pattern now validates per spec and invalid input fails closed.

### Verification

- `pytest -q` → **2553 passed, 3 skipped, 0 failed** (2536 baseline + 17)
- `ruff check`, `ruff format --check src/ tests/`, `mypy src` — clean
- **RED proof**: reverting only the two behavioral regexes (keeping the new function so collection still works — my first attempt at this proof failed on an ImportError and proved nothing) fails exactly the two new behavioral tests.

### Standing residual

`_semver_key` is hand-written, which is what I argued against when replacing the digit-extraction parser. It is now pinned against the spec's own §11.4 precedence chain in both directions, including `beta.2 < beta.11`, plus the rule 2/9/10 validation above. If a maintainer would rather take a dependency on a real semver library than keep this, that is a reasonable call — but it is no longer unvalidated.

Issue #79, opened 2026-06-28
Playwright/browser sessions unstable via gateway: long calls expire/timeout, refresh drops all servers, no browser reaping


ViperJuice:
## Summary

Driving the Playwright MCP server through the PMCP gateway (`mcp__pmcp__gateway_invoke` → `playwright::*`) from Claude Code, browser sessions are unstable in three distinct ways. **Short calls work flawlessly; long calls and recovery are broken.** This blocked automated UI verification of a deployed app.

**Environment:** Claude Code · PMCP gateway · `@playwright/mcp` (Chrome, headless, shared profile `~/.cache/ms-playwright-mcp/mcp-chrome-21048fe`) · observed 2026-06-28.

## Symptom 1 — long downstream calls are killed (the core issue)

Every **sub-second** discrete call succeeds reliably: `browser_navigate`, `browser_snapshot`, `browser_click`, `browser_take_screenshot`, `browser_evaluate`. But `browser_run_code_unsafe` running a multi-step loop (~3–10s — e.g. clicking through ~17 onboarding steps) fails in two ways:

- `"MCP server pmcp session expired"` — typically on the 2nd+ such call.
- `E303 "Tool execution timed out"` — even with `timeout_ms: 30000`.

→ Points to a **proxy/session timeout or missing keepalive** in the gateway's downstream-request layer that can't tolerate legitimately-long browser operations.

## Symptom 2 — `refresh` is destructive and doesn't reconnect

A single `gateway_refresh` reset **all 105 servers to 0 online** (`servers_seen: 105, servers_online: 0, tools_indexed: 0`), and none reconnected until a manual `gateway_provision('playwright')`. Expected: refresh restores the previously-running set.

## Symptom 3 — no browser reaping on session drop

When a long call drops the session, the Playwright server's Chrome is left **orphaned** (confirmed: live `chrome` procs on the profile after the drop), holding the profile's `SingletonLock`. The next launch then fails:

```
Error: Browser is already in use for …/mcp-chrome-21048fe, use --isolated to run multiple instances of the same browser
```

Manual recovery required killing the procs + deleting `Singleton*`.

## Asks

1. Raise (or heartbeat/keepalive) the downstream-call timeout for long browser ops — or stream progress so the session isn't torn down mid-call.
2. Make `gateway_refresh` reconnect previously-running servers instead of dropping to zero.
3. Reap the downstream browser (and clear its lock) on session end / drop.
[...]

PR #161, opened 2026-08-20
fix: update_server verifies config after the probe (#151)

### LINKED ISSUE #151: update_server resolves config twice across the probe window (TOCTOU)

ViperJuice:
Closes #151.

`gateway.update_server` resolved the target's config, ran a 60-second probe, then called `gateway.restart_server` — which resolved the config a **second, independent time**. The loaders re-read from disk on every call, so an edit landing in that window could make pmcp probe and install package A, restart onto package B, and record A's version as B's.

Now: resolve #1 → probe → resolve #2 → verify it still describes the same downstream process → restart onto **that exact verified object**. On mismatch, or if the server has vanished from config, refuse — fetched but **not** activated, **no** version recorded.

**Contract:** *the config restarted onto is the config that was probed.* Not "the config on disk when the restart completes" — an edit landing after the check applies on the next update. That is correct behaviour, not a residual race.

## The board caught a real defect in my first version

I put the guard immediately before the restart. Two reviewers independently blocked it, and they were right: between the probe and the restart, `update_server` calls `await self.refresh(...)` — and **`refresh()` is itself a diff-based config reconcile** that disconnects and reconnects servers whose definition changed. It could therefore **activate the freshly-fetched package on its own**, after which my guard would report "fetched but NOT activated" — a false statement.

The sharper half: `refresh()` activates via `disconnect_server` + `connect_all`, **not** `restart_server`. So my planned assertion — "`_client_manager.restart_server` was never called" — **would have passed while the bug survived.** That is the exact failure mode this repo has bitten me with repeatedly.

The guard now sits immediately after the probe and **before** `refresh()`, and the test asserts `refresh` was never called.

## A second correction, found while implementing

My plan specified a strict full-model `==` including `source`. That was wrong, and `handlers.py` already knew it: `_refresh_config_unchanged` is a battle-tested predicate for exactly this question, and it deliberately avoids full-model equality —

- **`env` is compared by *effective* override only.** Naive full-`env` comparison "would spuriously tear down running provisioned servers on every refresh (#79)". In `update_server` that class returns as **spurious refusals of legitimate updates**.
- **`source` is excluded deliberately** — the same effective server can arrive as `manifest` or as a configured entry.

I'd called over-refusal "rare and fails closed". The codebase had already settled the question with a linked issue. Reusing the predicate also keeps one definition of "same server" instead of two subtly different ones.

## Also corrected: the issue's own fix sketch

#151 says threading the config through would lose "policy checks, credential-gap detection, and remote-header validation". Verified against source — **all three live inside `_resolve_lifecycle_config`** (`handlers.py:3237-3295`), which read #1 already runs. `restart_server` adds only task accounting and output shaping. That is why extracting `_restart_resolved_server` is safe and loses nothing.

## Verification

- `pytest -q` → **2550 passed, 3 skipped, 0 failed** (2548 baseline + 2 new)
- `ruff check .`, `ruff format --check`, `mypy src` — clean
- **No existing test modified.** All 13 `update_server`/`restart_server` tests pass unchanged, which is what makes the helper extraction provably inert.
- **Tests proven RED three ways**, not assumed: against unfixed `c326a4c`; against my own board-blocked ordering (fails on `refresh_called`, the assertion that would otherwise have been vacuous); green against the fix.
- Prose swept — no comment, docstring, or exported tool description still claims the config is resolved once. The false comment the issue names is rewritten.

## Not attempted

Locking `.mcp.json` (pmcp does not own the file, and an operator's editor will not honour an advisory lock) or making the restart atomic with respect to disk (impossible without owning it).

---
<a href="https://app.blacksmith.sh/Consiliency/codesmith/pmcp/pr/161"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789827487&installation_model_id=427051&pr_number=161&repository=Consiliency%2Fpmcp&return_to=https%3A%2F%2Fgithub.com%2FConsiliency%2Fpmcp%2Fpull%2F161&signature=94d3c474fc78dccad06ef0f2006f48094111d06cb9ed14092171d08a73d020c7"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/.consiliency/plans/detailed-update-server-toctou-20260820-1300.md
+++ b/.consiliency/plans/detailed-update-server-toctou-20260820-1300.md
@@ -41,10 +41,23 @@ It is explicitly **not** "the config on disk at the instant the restart complete
 
 - `_restart_resolved_server` — **add** — private helper holding the post-resolve body of `restart_server`: active-task accounting, the `_client_manager.restart_server(config, force=...)` call, and `_lifecycle_output` shaping. Signature `(self, server_name: str, config: ResolvedServerConfig, *, force: bool, prior_status: str) -> LifecycleServerOutput`.
 - `restart_server` — **modify** — becomes resolve + delegate to the helper. **Behaviour must be byte-identical**; this is a pure extraction and any observable change is a bug in the refactor.
+**Board correction — the check must precede `refresh()`, not merely precede the restart.** My first implementation placed the guard immediately before the restart, which two reviewers independently blocked. Between the probe and the restart, `update_server` calls `await self.refresh(...)` — and `refresh()` is itself a **diff-based config reconcile**: it re-reads the config and disconnects/reconnects any server whose definition changed. So a mid-probe edit could cause `refresh()` to **activate the freshly-fetched package on its own**, after which the guard would report "fetched but NOT activated" — a false statement.
+
+Worse, it defeats the test I had planned. `refresh()` activates through `disconnect_server` + `connect_all`, **not** `restart_server`, so the planned assertion "`_client_manager.restart_server` was never called" would have passed while the bug survived. That is the exact failure mode this plan's test section warns about, and I walked into it.
+
+The guard therefore sits immediately after the probe's success check and **before** `refresh()`. The tests must assert the target's actual connect/disconnect state, not just that `restart_server` was skipped.
+
 - `update_server` — **modify** — after the probe returns:
   1. **Recompute `prior_status`.** The pre-probe value is up to 60 seconds stale. Use the fresh value for resolve#2 and the helper.
   2. Resolve #2 via `_resolve_lifecycle_config`. On failure, return `UpdateServerOutput(ok=False, ...)` exactly as the pre-probe failure path does. This also covers *server deleted from config mid-probe* for free.
-  3. Compare resolve#2's `ResolvedServerConfig` against resolve#1's with a **strict full-model `==`, including `source`**. A benign `source` flip (e.g. the same config moving project→user) will refuse; that is rare and fails closed, which is the correct bias here.
+  3. Compare resolve#2 against resolve#1 using the **existing `_refresh_config_unchanged(old, new)`** (`handlers.py:177`).
+
+     **Correction to this plan, found while implementing — the original said "strict full-model `==`, including `source`", and that was wrong.** `handlers.py` already carries a battle-tested predicate for exactly this question ("do these two describe the same downstream process?"), and it deliberately does *not* use full-model equality. Its docstring records why, and both reasons apply here:
+
+     * **`env` is compared by *effective* override only** — entries that genuinely differ from `os.environ`. Naive full-`env` comparison "would spuriously tear down running provisioned servers on every refresh (issue #79)". A full-model `==` in `update_server` would reintroduce that class as **spurious refusals of legitimate updates**.
+     * **`source` is excluded deliberately** — the same effective server can present as `manifest` or as a configured source. My original plan called over-refusal on a source flip "rare and fails closed". That was a judgement made without reading this function; the codebase had already decided the question, with a linked issue.
+
+     Reusing it also keeps **one** definition of "same server" instead of a second, subtly different one. `update_server` only reaches the comparison for local servers (remote returns early — no local package to update), so only the `LocalMcpServerConfig` branch is exercised; the header-rotation logic is inert here and needs no arguments.
   4. On mismatch — refuse. Return `ok=False` with a message saying the configuration changed during the update and the package was **fetched but not activated**, and do **not** restart, do **not** persist any version bookkeeping.
   5. On match — call `_restart_resolved_server` with **resolve#2's config object**. Since it is equal to #1 the choice is immaterial by construction; using the compared object is what makes the window exactly zero rather than one event-loop tick.
 - The comment block at `4924-4936` — **rewrite**. It currently asserts *"Resolving once through restart_server's own resolver makes that divergence impossible: both calls are the same lookup against the same inputs, so they can't disagree."* The premise "same inputs" is exactly what a mid-probe edit breaks. The issue names this comment specifically.
--- a/.consiliency/plans/detailed-update-server-toctou-20260820-1300.md
+++ b/.consiliency/plans/detailed-update-server-toctou-20260820-1300.md
@@ -1,0 +1,124 @@
+# Detailed plan: close the update_server TOCTOU window
+
+## Task
+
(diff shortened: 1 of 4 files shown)


ViperJuice:
## Board review: 2 AGREE / 1 DISAGREE — the dissent found a real hole, fixed in `675f180`

codex identified a case my guard let through, and I confirmed it empirically before changing anything rather than reasoning about it.

### The hole

`_refresh_config_unchanged` discards an explicit `env` entry whose value equals `os.environ` — correct for `refresh()`, which uses it to decide whether to tear down a running server. But `sanitized_subprocess_env` strips **every PMCP-managed key** and restores only explicit entries. Those two rules disagree:

```
managed keys: [... 'SECRET_TOKEN']
explicit env={"SECRET_TOKEN": "value"}  -> child has SECRET_TOKEN='value'
env=None                               -> child has NO SECRET_TOKEN
SPAWNED ENVS DIFFER:     True
COMPARATOR SAYS UNCHANGED: True
```

So an edit dropping that entry mid-probe passed the guard while the restarted server silently **lost the credential** — and the reverse edit newly **exposes** a managed credential to it. Same "restart onto something other than what was probed" defect, on a different axis. My existing test only changed argv and could not see it.

### The fix

The guard now also compares the **actual sanitized child environment**. Both sides are sanitized in one synchronous block, so ambient `os.environ` and env-store state are identical across the two calls and cancel out — only the config-driven difference survives, with no spurious-refusal risk from ambient drift.

I kept `_refresh_config_unchanged` for command/args/cwd. This is **not** a defect in it for `refresh()`'s purpose; the stakes differ here, where the question is whether the thing we probed is the thing we are about to launch.

### Verification

- `pytest -q` → **2551 passed, 3 skipped, 0 failed**
- ruff, format, mypy clean
- New regression test proven **RED against the env-blind comparator**
- The two earlier tests remain proven RED three ways: against unfixed `c326a4c`, against my own board-blocked ordering (fails on `refresh_called`), and green against the fix
- No existing test modified — all 13 `update_server`/`restart_server` tests still pass unchanged

### Verified for the other review questions

- **"fetched but NOT activated, no version recorded" is true on every path that can reach it.** The guard returns before `refresh()`, before `get_package_version`, and before the restart; version persistence sits behind `if restart_result.ok:`, which a refusal never reaches.
- **Awaits after the guard** are only `refresh()` and `get_package_version()`. The restart consumes the verified object rather than a third resolve, so the outcome stays consistent with what was probed.

ViperJuice:
## Re-review (third pass): 2 AGREE / 1 DISAGREE — split into a fix and a scope correction

I asked for this pass specifically because the env fix was written reactively to a dissent and had not itself been reviewed. It found two more things.

### 1. Real defect, fixed in `c3905f8` — spurious-refusal window

I claimed both sides were sanitized against "identical" env-store state. They were not. Each `sanitized_subprocess_env` call independently re-reads the credential files (`managed_secret_keys` hits disk per call), so a write landing between the two reads could make **identical configs compare unequal** — a spurious refusal on the only update path pmcp has. Two reads where I asserted one.

Both sides now derive from a **single** stripped base. Verified equivalent to `sanitized_subprocess_env` across every env shape (absent / empty / matching-ambient / differing / new key / PATH override).

### 2. Scope overclaim, corrected — not weakened

The reviewer is right that ambient `os.environ` and secret-store changes cancel out by construction, and that the spawn rebuilds the environment again in `ClientManager`. But that is what the check is *for*: isolating **config-driven** difference. The defect was my wording — I described it as covering "the spawned process environment", which is stronger than what it does.

The comment and CHANGELOG now say **config-driven**, and state plainly that the ambient environment is not frozen across the update.

Freezing it properly means threading a verified environment through `restart_server` into the spawn path — a wider change than this TOCTOU warranted, and one where the right behaviour is genuinely arguable (if an operator rotates a credential mid-update, should the restart pin the stale value or pick up the new one?). Filed as **#162** with that decision spelled out, rather than resolved by assumption here.

### Also in this push

The credential test is now parametrized over **both directions** — dropping an entry removes a managed credential, adding one newly exposes it. The reviewer asked whether the test covered the reverse; it did not. Both directions proven RED against the env-blind comparator.

### Verification

- `pytest -q` → **2552 passed, 3 skipped, 0 failed**
- ruff, format, mypy clean
- Every test in this PR proven RED against the specific bug it targets: unfixed `c326a4c`, my own board-blocked ordering (fails on `refresh_called`), and the env-blind comparator (both directions)
- No existing test modified — all 13 `update_server`/`restart_server` tests still pass unchanged

### Running tally

Three review passes, three findings, all of which I would have shipped: the guard placed after `refresh()`, the managed-credential env axis, and the double credential-store read. Recording that plainly because it is the honest summary of this PR.

Issue #158, opened 2026-08-19
Flaky: test_ec_p2_7_reconnect_does_not_leak_transports fails intermittently across Python lanes


ViperJuice:
`tests/runtime/test_downstream_remote.py::test_ec_p2_7_reconnect_does_not_leak_transports` fails intermittently in CI. Filed while landing #157, which it repeatedly blocked without being related to it.

## Evidence that it is order- or timing-dependent, not version-specific

Across consecutive runs of the **same branch**, the failure moved between Python lanes:

| run | 3.10 | 3.11 | 3.12 |
|---|---|---|---|
| first  | **FAIL** | pass | pass |
| second | pass | **FAIL** | pass |
| re-run of the second | pass | pass | pass |

A genuine version-specific bug would not migrate between lanes, and a re-run with **no code change** cleared it. It also fails at ~3% of the suite, early enough that alphabetically-later files are not the trigger.

It did not reproduce locally: passed in isolation and alongside `tests/test_tools.py` on 3.10, 3.11 and 3.12, and in several full-suite runs.

## Why it is probably fragile by construction

The module's own docstring explains that it asserts on **this process's open socket count** and that it must run in-process. It also documents a prior encounter with `sse_starlette.sse.AppStatus.should_exit` — a **process-global** class attribute latched `True` the first time any uvicorn server in the process shuts down, never reset.

A test that measures process-global socket counts is sensitive to anything else in the same interpreter holding a socket at the sampling moment, including timing differences between CI runners.

## What was ruled out

- **Not caused by #157.** I initially suspected my own new test, which spawns real process trees — a reasonable theory, since it had genuinely perturbed state. I moved that test into a child interpreter (`50a7f5c`) so it cannot touch shared state, and the failure still occurred, on a different lane.
- **Not the same as main's recent red runs.** Those were infrastructure: `429 Too Many Requests` downloading `astral-sh/setup-uv`. Worth stating because "main is also failing" is true but for an unrelated reason, and I nearly conflated the two.

## Suggested next step

Instrument the assertion to report *which* file descriptors are open when it fails, rather than only the count. Without that, each occurrence gives one bit of information and the loop repeats.

ViperJuice:
Fixed in #163 (merged). Root cause found, not worked around.

## The cause

[...]

PR #163, opened 2026-08-21
fix: close the four #155 follow-ups (#156)

### LINKED ISSUE #156: Follow-ups from #155: hand-written semver, negation hazard, digest ambiguity, CI guard fragility
Known gaps left open when #155 (fail-closed version comparison) merged. None block that PR — it is strictly better than the digit-extraction it replaces — but each is real and worth recording rather than rediscovering.

## 1. `_semver_key` is hand-written

#155 replaced a hand-rolled parser with `packaging.version` precisely because hand-rolled version logic had produced a fabricated-notice bug in every form it took. The npm/Cargo path then had to be hand-written anyway, because PEP 440 and SemVer genuinely disagree: PEP 440 reads `1.0.0-1` as the post-release `1.0.0.post1`, while SemVer reads it as a prerelease *below* `1.0.0`. 79 of the manifest's 107 servers are npm, so that inversion was live.

It is now validated rather than assumed — pinned against the SemVer 2.0.0 §11.4 canonical precedence chain in both directions (including `beta.2 < beta.11`, which naive string ordering reverses), plus rule 2/9/10 validation rejecting leading zeros and empty identifiers. But it remains code we maintain instead of a library.

**Option:** take a dependency on a real semver package and delete `_semver_key`. Trade-off is one more runtime dependency against removing the last hand-rolled comparison.

## 2. `is_version_orderable` exists because a negated caller silently inverted

Making `is_version_newer` fail closed changed the meaning of every `not is_version_newer(...)` in the codebase. `refresher.py` treats "not newer" as "up to date", so once unreadable input started returning `False`, an unorderable cached version (the literal `"unknown"` that same function persists after a failed lookup) read as *current* — pinning the cache forever.

Measured:

```
main:     is_version_newer('unknown','2.0.0') -> True   =>  not(...) = False  => regenerates  (correct)
after:    is_version_newer('unknown','2.0.0') -> False  =>  not(...) = True   => stale forever (broken)
```

Fixed by gating that call site on `is_version_orderable()`. **The generalisable lesson: I audited the eight call sites for arity but not for negation.** Any future change to a shared predicate's failure semantics needs a negation audit, not just a signature audit.

**Gap:** nothing prevents a *new* negated caller from reintroducing this. A lint rule or a comment on `is_version_newer` warning that `not` is unsafe without an orderability check would help.

## 3. Digest detection is shape-based when the package type is unknown

`get_docker_version` truncates SHA-256 to 12 hex chars, which can be all digits (`987654321098`). A calendar version like `202612180000` has the same shape. With `package_type == "docker"` the type disambiguates; without it, a hex letter is required, so an **all-numeric digest arriving without a type is treated as a version**.

Reachable only if a caller loses the package type. All current callers pass it, and the memo now carries it, so this is latent rather than live.

## 4. The changelog CI guard called an endpoint its token cannot reach

Found while merging #155. The `changelog` job grants `pull-requests: read` but called `/issues/<n>/labels`, which needs `issues: read` — GitHub returned `Validation Failed (HTTP 422)`, and under `bash -e` that killed the step *before* the CHANGELOG check ran. The guard blocked the PR on its own permissions rather than on its content.

Fixed in #155 (switched to `/pulls/<n>` `.labels`, and made the lookup tolerate failure with `|| true` so an API hiccup means "no label seen" rather than "fail the check"). Recorded here because it went unnoticed since the guard shipped: it only fires on a PR that touches `src/` *and* reaches the label branch.

**Residual:** the guard is still one `gh` call away from a false block. Worth considering whether it should fall back to the frozen event payload when the live lookup fails.

## Context

#155 was reviewed in four rounds; each found real defects, and two of them were damage introduced while fixing the previous round. The full history is on that PR. The underlying issue #150 (wrong "update available" notices) remains open and needs a version source whose identity is bound to the package being compared — five attempts are documented across #154 and two plan artifacts.

ViperJuice:
Closes #156 — all four recorded follow-ups from #155.

## 1. `_semver_key` replaced by the `semver` library

#155 moved this module to `packaging` precisely because hand-rolled version logic produced a fabricated-notice bug in every form it took — then had to hand-roll the SemVer lane anyway, because PEP 440 and SemVer genuinely disagree (`1.0.0-1` is a *post*-release to PEP 440, a *prerelease* to SemVer, and 79 of the manifest's 107 servers are npm).

`semver` 3.x is pure-Python with **zero transitive dependencies**. Before swapping, I verified it rejects every case the hand-written regex existed to reject — `01.0.0`, `1.0.0-01`, `1.0.0-a..b`, `1.0`, `build-1`, `nightly`, `1.0.0.post1` — and orders `beta.2 < beta.11`, which naive string comparison reverses.

**All 85 existing version-checker tests pass unmodified**, including the SemVer 2.0.0 §11.4 precedence chain #155 pinned in both directions. That is what makes this a swap rather than a rewrite.

## 2. Negation hazard now enforced, not just documented

`is_version_newer` fails closed, so `False` means *either* "current" *or* "unorderable". A bare `not is_version_newer(...)` collapses those — which is exactly how making it fail closed silently inverted `refresher.py` and pinned a stale cache forever.

Added an explicit warning to the docstring, plus `test_no_unguarded_negation_of_is_version_newer`: an **AST** walk (not grep — comments and strings don't register) that fails if any `not is_version_newer(...)` appears without an `is_version_orderable(...)` guard in the same statement.

**Proven to fire**: removing the real guard from `refresher.py` fails the test with the offending file:line.

## 3. Digest ambiguity — and a real bug found on the way

**I tried fail-closed first and reverted it.** Refusing to order any bare 12-digit string does close the ambiguity, but it breaks CalVer (`202612180000`) — which #155 deliberately supports and pins with `test_long_numeric_version_is_not_mistaken_for_a_digest`. That existing test caught me trading a **live capability** for a **latent** bug the issue itself calls "not live". The test was right.

The partner value disambiguates instead: the two values are compared as a pair from one server, so if either side is unmistakably a digest (a hex letter, or the `sha256:` prefix), the all-numeric side is one too. A truncated digest is all-numeric roughly 1 time in 11,500, so the mixed pair covers essentially the whole real surface, and CalVer keeps working.

**Separate bug this surfaced:** the digest pattern required a hex letter *even when `sha256:` was present*, so `sha256:987654321098` — unambiguously a digest — was rejected outright. The prefix now settles it; the letter requirement applies only to bare hex.

## 4. CHANGELOG CI guard

`|| true` stopped a failed lookup from killing the step, but left "lookup failed" and "no label" indistinguishable — so an API hiccup **false-blocks a PR that really was labelled**. It now falls back to the frozen event payload before concluding the label is absent.

The payload is passed via `env:` rather than interpolated into the script body, so a label's text can never be parsed as shell. Fallback logic tested locally across eight cases (live wins over payload, payload used only when live is empty, and malformed/`null` JSON degrade to "enforce" without crashing).

## Verification

- `pytest -q` → **2557 passed, 3 skipped, 0 failed**
- `ruff check .`, `ruff format --check`, `mypy src` — clean
- **Fresh-install smoke**: built the wheel, installed into a clean venv, confirmed `semver` resolves and npm prerelease ordering works. `mypy` cannot catch a missing runtime dep here (`ignore_missing_imports = true`), and ruff doesn't resolve imports — I only caught the undeclared dependency by checking. Locked in `uv.lock`.
- Every new test proven RED against the specific defect it targets.

---
<a href="https://app.blacksmith.sh/Consiliency/codesmith/pmcp/pr/163"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789878200&installation_model_id=427051&pr_number=163&repository=Consiliency%2Fpmcp&return_to=https%3A%2F%2Fgithub.com%2FConsiliency%2Fpmcp%2Fpull%2F163&signature=66f10b11203e7d788726b38004a94cf7311343518186984df8a77d42c0e6ca47"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -213,6 +213,10 @@ jobs:
       - name: Require a CHANGELOG.md entry for src/ changes
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          # Frozen at trigger time -- the fallback below, not the primary source.
+          # Passed as env rather than interpolated into the script body so a
+          # label's text can never be parsed as shell.
+          PAYLOAD_LABELS_JSON: ${{ toJSON(github.event.pull_request.labels.*.name) }}
         run: |
           # Query live labels via the API instead of github.event.pull_request.labels.
           # That context is frozen in the event payload at the moment the workflow
@@ -225,9 +229,18 @@ jobs:
           # so it returned "Validation Failed (HTTP 422)". Under `bash -e` that
           # non-zero exit killed the step BEFORE the CHANGELOG check ran, so the
           # guard failed on its own permissions rather than on the PR's content.
-          # `|| true` keeps a future API hiccup from doing the same: a lookup
-          # failure must not decide the check, it just means "no label seen".
+          # `|| true` keeps a future API hiccup from killing the step. But
+          # "lookup failed" is NOT "no label": treating them alike false-blocks
+          # a PR that really was labelled (Consiliency/pmcp#156 item 4). On an
+          # empty result, fall back to the frozen event payload -- stale for a
+          # label applied after the trigger, but strictly better than ignoring
+          # the label outright. Live lookup first, payload second, "no label"
+          # only if both come up empty.
           LABELS=$(gh api "repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}" --jq '.labels[].name' || true)
+          if [ -z "$LABELS" ]; then
+            echo "Live label lookup empty or failed — falling back to the event payload."
+            LABELS=$(printf '%s' "$PAYLOAD_LABELS_JSON" | jq -r '.[]?' 2>/dev/null || true)
+          fi
           echo "Current labels: $LABELS"
           if echo "$LABELS" | grep -qx 'skip-changelog'; then
             echo "skip-changelog label present — skipping the CHANGELOG.md check."
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -229,13 +229,10 @@ jobs:
           # so it returned "Validation Failed (HTTP 422)". Under `bash -e` that
           # non-zero exit killed the step BEFORE the CHANGELOG check ran, so the
           # guard failed on its own permissions rather than on the PR's content.
-          # `|| true` keeps a future API hiccup from killing the step. But
-          # "lookup failed" is NOT "no label": treating them alike false-blocks
-          # a PR that really was labelled (Consiliency/pmcp#156 item 4). On an
-          # empty result, fall back to the frozen event payload -- stale for a
-          # label applied after the trigger, but strictly better than ignoring
-          # the label outright. Live lookup first, payload second, "no label"
-          # only if both come up empty.
+          # A lookup FAILURE is not "no label": treating them alike false-blocks
+          # a PR that really was labelled (Consiliency/pmcp#156 item 4). On
+          # failure, fall back to the frozen event payload -- stale for a label
+          # applied after the trigger, but better than ignoring the label.
           # Branch on the EXIT STATUS, never on emptiness. A successful lookup
           # returning zero labels is authoritative -- it means the label was
           # removed. Falling back to the frozen payload there would resurrect a
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -236,9 +236,16 @@ jobs:
           # label applied after the trigger, but strictly better than ignoring
           # the label outright. Live lookup first, payload second, "no label"
           # only if both come up empty.
-          LABELS=$(gh api "repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}" --jq '.labels[].name' || true)
-          if [ -z "$LABELS" ]; then
-            echo "Live label lookup empty or failed — falling back to the event payload."
+          # Branch on the EXIT STATUS, never on emptiness. A successful lookup
+          # returning zero labels is authoritative -- it means the label was
+          # removed. Falling back to the frozen payload there would resurrect a
+          # deleted `skip-changelog` and silently bypass this guard, which is
+          # worse than the false block the fallback exists to prevent
+          # (ah board review).
+          if LABELS=$(gh api "repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}" --jq '.labels[].name'); then
+            echo "Live label lookup succeeded."
+          else
+            echo "Live label lookup FAILED — falling back to the event payload."
             LABELS=$(printf '%s' "$PAYLOAD_LABELS_JSON" | jq -r '.[]?' 2>/dev/null || true)
           fi
           echo "Current labels: $LABELS"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,8 +24,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   version reading as current — but it checked only the cached side. With a
   cached `1.0.0` and a fetched `nightly`, the guard passed, the comparison
   failed closed, and the negation still reported "up to date", never
-  refreshing. Both sides are now checked, and both checks carry the package
-  type.
+  refreshing.
+
+  Checking each side individually turned out to be insufficient too:
+  comparability is a property of the **pair**. `1.0.0` and `abcdef123456` are
+  each orderable on their own, but a version and a digest cannot be ordered
+  against each other, so the same "up to date" answer came back for a server
+  whose cache entry was reused by name after its package type changed. A new
+  `are_versions_comparable(current, latest, package_type)` asks about the pair,
+  and that is what now guards the short-circuit.
 - **An all-numeric truncated image digest is documented as incomparable
   without a package type, and callers are now pinned to pass one.**
   `get_docker_version` truncates SHA-256 to 12 hex characters, which can be all
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,12 +42,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   isolation. The flag is now cleared on entry as well as exit, with a test that
   latches it deliberately and asserts a connection still works.
 - **The CHANGELOG CI guard no longer treats a failed label lookup as "no
-  label".** A live-lookup failure now falls back to the frozen event payload
+  label".** A live-lookup *failure* now falls back to the frozen event payload
   before concluding the `skip-changelog` label is absent, so an API hiccup
-  cannot block a PR that really was labelled.
+  cannot block a PR that really was labelled. A lookup that *succeeds* and
+  returns no labels stays authoritative — otherwise removing the label would
+  not re-enable the check.
 
-
-### Fixed
 - **`gateway.update_server` no longer risks restarting onto a different package
   than the one it probed.** The tool resolved the server's config, ran an update
   probe with a 60-second timeout, and then let `gateway.restart_server` resolve
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,14 +18,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   SemVer lane was the last piece still hand-written.
 
 ### Fixed
-- **An all-numeric truncated image digest is no longer misread as a version.**
+- **An all-numeric truncated image digest is documented as incomparable
+  without a package type, and callers are now pinned to pass one.**
   `get_docker_version` truncates SHA-256 to 12 hex characters, which can be all
-  digits — the same shape as a calendar version like `202612180000` — so with
-  no package type the shape alone could not classify it, and a genuinely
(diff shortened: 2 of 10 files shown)


ViperJuice:
## Also closes #158 — root cause found, not worked around

CI failed twice on this branch with `test_ec_p2_7_reconnect_does_not_leak_transports`. That test is unrelated to anything here (version comparison, CI workflow), so I checked rather than assumed: **run `32369319515` on `main` at `7318f7e` failed with the identical test.** Not caused by this PR.

Since it was blocking the merge, I traced it.

### Root cause, confirmed empirically

```
latched=False: connect OK
latched=True:  FAILED -> Exception: SSE stream ended without a response
```

The exact error CI reported. `sse_starlette.sse.AppStatus.should_exit` is a **process-global class attribute** that uvicorn's shutdown handler latches `True` and never resets. `run_fake_remote` cleared it in its `finally` block — which protects against its *own* shutdown, but does nothing about a server stopped elsewhere in the same interpreter. `tests/mcp2x/test_listen_over_http.py` and `tests/test_http_dos.py` both stop uvicorn servers, and **`tests/mcp2x` sorts immediately before `tests/runtime`**.

That explains both halves of the puzzle #158 recorded: why it failed at ~3% of the suite, and why it never reproduced in isolation.

### The fix

Clear the flag on **entry** as well as exit, so the server is immune to whatever ran before it. Plus a test that latches the flag deliberately — the state a real full-suite run produces — and asserts a connection still works. **Proven RED without the entry reset.**

Validated in the actual CI ordering too: `tests/mcp2x tests/runtime tests/test_http_dos.py` in one process, 143 passed.

### A note on #158's suggested next step

The issue proposed instrumenting the assertion to report which file descriptors were open. That was the right instinct for a socket-count failure — but the failures were the **SSE stream dying**, not the count. The socket assertion never got the chance to run. Worth recording, since it's why the earlier investigation stalled.

### Verification

- `pytest -q` → **2558 passed, 3 skipped, 0 failed**
- ruff, format, mypy clean
- Every test in this PR proven RED against the specific defect it targets

ViperJuice:
## Board review: 2 AGREE / 1 DISAGREE — three findings, all valid, all fixed in `075a18c`

### 1. Partner promotion fabricated updates — reverted

I flagged this exact case myself and judged it defensible. I was wrong, and the reviewer's reasoning is better: promotion resolves the ambiguity by **guessing**, and the guess fabricates. `is_version_newer("202612180000", "abcdef123456")` returned `True` — a genuine calendar version reported as an update because its partner looked like a digest. That is precisely what the fail-closed contract exists to prevent.

The deciding factor is that the case promotion fixed is **latent** (both live callers pass `pkg_type`), while the case it broke is a **live** contract violation. Same trade I got wrong in the other direction earlier in this PR, when I tried fail-closed and an existing CalVer test caught me.

Mixed pairs are now incomparable, with the reasoning recorded in-source so the next person doesn't re-litigate it. **The `sha256:`-prefix fix is retained** — that one was unambiguous, not a guess.

Replacing it: `test_all_is_version_newer_callers_pass_package_type` pins the thing that actually resolves the ambiguity. #156 called the risk latent *because* every caller passes the type; this makes that a guarantee instead of an observation that happened to hold. Proven to fire — dropping the type from one caller names `refresher.py:429`.

### 2. The AST test checked co-occurrence, not guarding — fixed

Correct, and worth stating plainly: **my first strengthening still accepted the reviewer's exact bypass.** Walking into each `and` operand finds a guard nested inside an `or` beside the negation, which short-circuits nothing.

It now requires the guard to be a **sibling** operand of the `and` chain, positioned **before** the negation, and about the **same** value. Verified against four bypasses:

| bypass | result |
|---|---|
| `not is_version_newer(a,b) or is_version_orderable(a)` | rejected |
| guard on the wrong subject | rejected |
| no guard at all | rejected |
| guard placed *after* the negation | rejected |

Real code still passes.

### 3. CI fallback conflated "failed" with "zero labels" — fixed

Also correct, and the worst of the three, because it made the guard *weaker* than before my change: a `skip-changelog` label that had been **removed** would be resurrected from the frozen payload and silently bypass the CHANGELOG check.

Now branches on **exit status**, never emptiness — a successful lookup returning zero labels is authoritative. Tested across five cases: label present, label removed, lookup failed with payload, lookup failed with empty payload, malformed JSON.

### Verification

- `pytest -q` → **2560 passed, 3 skipped, 0 failed**
- ruff, format, mypy clean
- Every test proven RED against the specific defect it targets

ViperJuice:
## Board round 3: the item-2 bug was still live. Fixed in `33f7961`.

This is the finding that mattered. After three rounds of me refining a *lint* to police the guard, the reviewer checked the **guard itself** — and it was wrong.

```
is_version_orderable('1.0.0')             = True     <- cached side, passes
is_version_newer('1.0.0','nightly','npm') = False    <- fails closed
guard and not newer                       = True     -> "up to date"
```

`is_version_orderable(existing_cache.version)` guarded only the **cached** side. An unreadable **fetched** version made the comparator fail closed, the negation read "up to date", and the stale cache pinned forever — **the exact defect #156 item 2 exists to eliminate**, surviving every round of this PR.

Both sides are now checked, and both checks carry `pkg_type` (the type is what makes an all-numeric docker digest orderable at all).

### Tests

- **Behavioural** regression test on the observable: `refresh_server` calls `get_package_version` a second time on the refresh path, so a short-circuit is detectable. Asserting the return value could not distinguish "refreshed" from "returned the stale cache". **Proven RED** against the one-sided guard.
- The lint now requires **both** operands guarded and names the unguarded one. Also proven RED — it reported `refresher.py:232 (unguarded: version)`.

### On the lint, honestly

The reviewer also showed it accepts a guarded value that is **reassigned** before use, and a guard whose **scope has ended** before a closure evaluates the comparison. That is correct: a syntactic check cannot prove a dataflow property.

I am not going to keep patching it. Its limits are now documented in the test itself, and **#164** proposes the structural fix: replace the fail-closed boolean with a tri-state `newer / not_newer / incomparable`, which cannot be collapsed by `not`. That makes the hazard unrepresentable and the lint unnecessary. Split out on the maintainer's call so this PR can land.

I also found and fixed two false positives myself while pre-attacking the review questions: a walrus-assigned guard, and (documented, not fixed) an early-exit guard.

### Score so far

Four rounds, and the reviewer was right every time — including about the thing I had already "fixed" three times. Recording that plainly.

### Verification

- `pytest -q` → **2561 passed, 3 skipped, 0 failed**
- ruff, format, mypy clean
- Every test proven RED against the specific defect it targets

ViperJuice:
## Board round 4: three findings — two fixed in `f77491c`, one filed as #165

### 1. Package identity → filed as #165

Correct, and verified: the short-circuit block never references `existing_cache.package`, and `pkg_name` is not resolved until after the early return. A cache for `old-package@1.0.0` reused for `new-package@1.0.0` returns the wrong descriptions, indefinitely.

Filed separately because it is a different question. This PR fixes whether two versions can be **ordered**; that one is whether they describe the **same thing**. The cross-ecosystem instance (npm cache vs docker config) *is* fixed here — a version and a digest are incomparable — but the same-ecosystem case is not, and it is pre-existing.

### 2. The lint ignored `package_type` — fixed

Correct and safety-relevant. `are_versions_comparable("202612180000", "1.0.0")` is True while the same pair classified as docker is incomparable, so a guard that dropped the type would pass the lint while the comparator beside it failed closed. All three operands must now match.

### 3. The drift test checked the wrong direction — fixed

The sharpest of the three. My test skipped every pair the predicate *accepts*, so it could only catch "predicate rejects something orderable" — which suppresses updates. The dangerous direction is the opposite: the predicate accepting a pair the comparator cannot order, which is exactly what lets `not is_version_newer(...)` report "up to date".

Now tested across the same 28-value × 7-type corpus: an accepted pair must either order in one direction, or be canonically the **same value**.

**Writing it exposed two bugs in my own canonical-identity helper**, both reporting drift that wasn't there:

- comparing rendered strings, so `1.0` vs `1.0.0` — the same PEP 440 release — looked different;
- then comparing full `Version` objects, so `1.0.0` vs `1.0.0+b` looked different, when `is_version_newer` deliberately compares `.public` because build metadata is not a new release.

Proven RED by injecting drift into `are_versions_comparable`: the test names `('1.0.0', 'abcdef123456', None)` as comparable-but-unorderable.

### Verification

- `pytest -q` → **2568 passed, 3 skipped, 0 failed**
- ruff, format, mypy clean
- I also independently swept 5488 pairs (28 values × 7 types) for safety violations before committing: zero.

### Tally

Five rounds, seven findings, every one valid — including two about defects I had already "fixed". Recording that plainly rather than burying it.

PR #170, opened 2026-08-23
feat: tri-state version comparison (TRISTATE, #164)

### LINKED ISSUE #164: Version comparison should be tri-state, not a fail-closed boolean
Surfaced during board review of #163. #163 fixes the live instance; this is the structural fix, deliberately deferred so that PR could land.

## The recurring shape

`is_version_newer(current, latest, package_type)` **fails closed**: it returns `False` both when *current* is genuinely up to date and when the pair could not be ordered at all. Two distinct meanings, one value.

Every consumer that writes `not is_version_newer(...)` collapses them into "up to date". That has now caused the same defect **three separate times**:

1. **#155** — making the comparator fail closed silently inverted `refresher.py`'s existing negated caller. A cached `"unknown"` began reading as current, pinning the cache forever.
2. **#156 item 2** — `is_version_orderable()` was added as a guard, and a lint was proposed to stop new negated callers appearing.
3. **#163 board review** — the guard itself was wrong: it checked only the *cached* side. With a cached `1.0.0` and a fetched `nightly`, the guard passes, the comparator fails closed, and the negation still reports "up to date". Fixed there by guarding both sides and passing `package_type` to both checks.

Three fixes, same root: **a boolean cannot express three outcomes.**

## Why the lint is not the answer

#163 added an AST test that rejects an unguarded `not is_version_newer(...)`. It took four attempts to get right, and two reviewers independently demonstrated bypasses. The committed version still models syntax rather than value flow — it accepts a guarded value that is *reassigned* before use, and a guard whose scope has ended before a closure evaluates the comparison. Its known limitations are documented in the test.

A syntactic check cannot soundly prove a dataflow property. It is a smoke alarm, not a guarantee.

## Proposal

Add a tri-state comparison as the primary API:

```python
def compare_versions(current, latest, package_type=None) -> Literal["newer", "not_newer", "incomparable"]:
```

`"incomparable"` cannot be collapsed by `not`. A caller must handle it explicitly or fail a type check, so the defect becomes unrepresentable rather than merely detectable.

Migration: keep `is_version_newer` as a thin wrapper (`compare_versions(...) == "newer"`) so nothing outside breaks, move both `refresher.py` call sites to the tri-state, and then **delete the AST lint** — it exists only to police a hazard the tri-state removes.

## Worth deciding

Whether `is_version_orderable` survives. It exists solely to make the negation safe; with a tri-state, "is this orderable" is answered by `compare_versions(...) != "incomparable"`, and keeping a second way to ask invites the two from drifting.

ViperJuice:
Executes **TRISTATE**, phase 1 of `specs/phase-plans-v12.md`, per `plans/phase-plan-v12-TRISTATE.md`. Closes Consiliency/pmcp#164.

## What changed

`is_version_newer` **failed closed** — `False` meant *either* "up to date" *or* "cannot be ordered". Every consumer writing `not is_version_newer(...)` collapsed those, and that exact defect was fixed three separate times (Consiliency/pmcp#155, #156, #163). The lint written to police it was bypassed by reviewers **four times**, because a syntactic check cannot prove a dataflow property.

Replaced by `compare_versions(current, latest, package_type) -> Literal["newer", "not_newer", "incomparable"]` as the sole classification path. **`is_version_newer` and `are_versions_comparable` are deleted outright**, not kept as wrappers — a function that does not exist cannot be negated. The AST lint goes with them.

`is_version_orderable` stays; it answers a genuinely unary question.

## Behaviour is unchanged — this is a representation change

All 12 parity cases from the plan verified against the merged tree:

| current | latest | type | result |
|---|---|---|---|
| `1.0.0-rc1` | `1.0.0` | npm | newer |
| `1.0.0-1` | `1.0.0` | npm | newer *(SemVer, not PEP 440)* |
| `1.0.0` | `1.0.0+b` | — | not_newer *(build metadata is not a release)* |
| `1.0` | `1.0.0` | — | not_newer *(same PEP 440 release)* |
| `abcdef123456` | `sha256:abcdef123456` | docker | not_newer *(same image, two spellings)* |
| `202612180000` | `202612190000` | — | newer *(CalVer)* |
| `1.0.0` | `abcdef123456` | docker | incomparable |
| `unknown` | `2.0.0` | npm | incomparable |

The existing `TestUpToDateShortCircuit` tests pass **unchanged** — the plan required that, on the grounds that needing to edit them would mean the migration changed behaviour.

## Execution

One working lane, worktree-isolated, tasks in order: RED tests → add `compare_versions` → migrate `refresher.py` → delete the wrappers. The ordering is the safety mechanism; deleting before migrating breaks the tree with an `ImportError` that reads like a merge problem.

Pre-merge audits all clean: no parent-tree leakage, exactly the 4 owned files touched, no file deletions.

## Verification

- `pytest -q` → **2582 passed, 3 skipped, 0 failed** (baseline 2568; net +14 after new tri-state tests and deleted lint tests)
- Interface freeze IF-0-TRISTATE-1 verified closed: `compare_versions` present, both wrappers absent, `is_version_orderable` retained

## Still to come on this branch

The `SL-docs` lane — CHANGELOG entry and a roadmap amendment recording that the roadmap's 3-lane file split did not survive the wrapper-deletion decision. Draft until that lands.

---
<a href="https://app.blacksmith.sh/Consiliency/codesmith/pmcp/pr/170"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790055739&installation_model_id=427051&pr_number=170&repository=Consiliency%2Fpmcp&return_to=https%3A%2F%2Fgithub.com%2FConsiliency%2Fpmcp%2Fpull%2F170&signature=586bbf95dbcede17beaa0b1d330a13ee4131cdf7b86106ae2236db60fda047ae"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


ViperJuice:
## Phase complete — both lanes merged, ready for review

`SL-docs` landed: CHANGELOG entry, docs-catalog rescan, and a roadmap amendment recording that the roadmap's 3-lane file split did not survive the wrapper-deletion decision.

### The pre-merge audit earned its keep

`SL-docs` **leaked into the parent checkout** — its writes landed both in the worktree and in `/home/viperjuice/code/pmcp` directly. The Step 6.5 leakage check caught it before the merge.

The skill's guidance is to confirm the leak is a byte-identical redundant copy and then discard it. `CHANGELOG.md` was identical. **`specs/phase-plans-v12.md` diverged by 13 lines**, so I inspected it instead of discarding blind — and the parent copy had silently deleted the `### Phase 2 — Update-path identity and environment contracts (UPDPATH)` heading.

Confirmed by the validator rather than by eye:

```
parent (leaked, uncommitted):  validate_roadmap: OK — 2 phase(s)
lane's committed version:      validate_roadmap: OK — 3 phase(s)
```

Had that copy won, the roadmap would have lost UPDPATH entirely and still passed validation — a silent one-phase deletion. The lane's commit was correct; the parent copy was a damaged intermediate. Discarded it, merged the lane, and re-validated on the merge target: **3 phases**.

### Verification

- `pytest -q` → **2582 passed, 3 skipped, 0 failed** (baseline 2568)
- Roadmap validator on the merged tree: 3 phases
- All 5 acceptance criteria checked mechanically:
  - **EC-TRISTATE-1/3** — `compare_versions` present; `is_version_newer` and `are_versions_comparable` absent; `is_version_orderable` retained
  - **EC-TRISTATE-2** — short-circuit tests pass, incomparable pairs do not short-circuit
  - **EC-TRISTATE-4** — corpus and drift tests pass
  - **EC-TRISTATE-5** — CHANGELOG names `compare_versions`

### Execution notes

`SL-docs` needed one retry — the first attempt stalled before creating its worktree and did no work, so it was killed and respawned at the escalated tier rather than resumed. `SL-1` ran clean first time.

Worktrees pruned, lane branches deleted, tree clean.

ViperJuice:
## Follow-up: the docs lane's own report caught a regression I had already merged

The `SL-docs` lane's completion report arrived after I merged it, and it flagged something worth acting on: the docs-catalog `--rescan` dropped `SPEC_COMPLIANCE.md` while the file still exists. The lane ran the tool as instructed rather than second-guessing it, and reported the discrepancy — correct behaviour, and the only reason this was caught.

I verified it rather than taking it on faith:

```
catalog on main:   80 entries, includes SPEC_COMPLIANCE.md
after --rescan:    82 entries, SPEC_COMPLIANCE.md ABSENT
file on disk:      7.4 KB, tracked, still present
```

**Why it was easy to miss:** the entry count went *up*. The rescan correctly picked up four new files under `specs/` and `plans/`, so the diff reads as net growth.

**Why it matters:** the catalog is what every phase's `SL-docs` lane iterates to decide which documents a change affects. `SPEC_COMPLIANCE.md` would have been skipped by every future phase, silently.

Root cause is in the shared helper, not this repo — `_find_files` collects root-level docs only from a hardcoded `KNOWN_FILES` allowlist, so any root doc outside it is dropped on **every** rescan. Filed as Consiliency/pmcp#171 with two suggested fixes; the better one is to merge-rather-than-replace so a removal is visible instead of silent.

Restored the entry in `dde5168`. Catalog is back to 83 entries with `SPEC_COMPLIANCE.md` present.

## Also worth recording: the parent-tree leak has an explanation

The lane reported that its first edit pass "reported success from the Edit tool but silently did not persist to disk", and that re-applying with absolute worktree paths worked.

That is almost certainly not a silent no-op — it is the **parent-tree leak** I caught at Step 6.5. The first pass resolved relative paths against the parent checkout, so the writes landed in `/home/viperjuice/code/pmcp` rather than the worktree. That explains both halves: the lane saw no change in its worktree, and the parent ended up holding that first, broken pass — the one missing the UPDPATH phase heading.

So the leaked parent copy was the lane's *first attempt*, not a duplicate of its commit. The Step 6.5 byte-equality check is what stopped it from being discarded blind or, worse, surviving.

ViperJuice:
## Panel CR: 2 AGREE / 1 DISAGREE — two blocking findings, both real, both fixed

You were right to ask. I had boarded the roadmap and the plans but merged the **executed code** to the phase branch without a panel pass — the first PR in this stretch that touches `src/`. Boarded now with the four seats: grok-4.6, gpt-5.6-sol, gemini-3.7-flash-high, and fable filled natively.

### 1. EC-TRISTATE-1 was unmet — `compare_versions` was not the sole classification path

`is_version_orderable` re-derived the digest / SemVer / PEP-440 classification itself, so the private classifiers had **two** call sites that could drift apart. That is precisely the hazard 2.2.1's wrapper-drift corpus existed to police, reintroduced somewhere new — and the phase's whole premise is that one classification site is what makes the collapse unrepresentable.

I reached this one independently while pre-attacking the review questions, and sol flagged it as blocking.

Fixed by delegation: `compare_versions(value, value, t) != "incomparable"`. Comparing a value against itself is orderable exactly when the value is. Verified equivalent to the previous implementation across **112 value/package-type combinations** before replacing it. The private classifiers are now consulted from one function only.

### 2. EC-TRISTATE-4 was neither enforced nor correct

**Not enforced.** The corpus asserted only the safety direction (incomparable ⇒ ordered in neither direction) and never that an ordinal result is *antisymmetric*. sol claimed a mutant returning `"newer"` both ways would pass — I verified it, and it does:

```
real mutant (SemVer lane always "newer"):
  corpus + compare_versions tests → 7 passed   ← undetected
  new TestOrdinalReversal          → 2 failed  ← caught
```

Worth noting my own first mutant was a **no-op** — that branch is a ternary, so the naive string substitution changed nothing and I briefly read it as "the test is weak." The real mutant confirmed sol's claim exactly.

**Not correct.** The criterion demanded that *every* `newer` pair reverse to `not_newer`. That is false for the digest lane: a digest is an **identity**, not an ordinal, so two different digests are each "a new image" relative to the other and read `newer` in both directions.

```
1.0.0 → 2.0.0 (npm)                 fwd=newer  rev=not_newer   ordinal
abcdef123456 → abcdef123457 (docker) fwd=newer  rev=newer       identity
```

`TestOrdinalReversal` now asserts antisymmetry for the ordinal lanes and **pins the digest exemption explicitly**, so nobody later "fixes" it into antisymmetry. EC-TRISTATE-4 is amended in the roadmap with the reason recorded.

### Verification

- `pytest -q` → **2585 passed, 3 skipped, 0 failed**
- ruff, format, mypy clean; roadmap validator still 3 phases
- New tests proven RED against the specific mutant they target

ViperJuice:
## Fable seat: AGREE — and it verified the change far better than I did

Final seat reported. It reviewed as the seat that argued for deletion over policing at planning time, judging whether the code cashes that argument.

**Its method beats mine.** I checked 12 hand-picked parity cases. It extracted `main`'s `version_checker.py`, loaded it alongside the branch's, and ran a **brute-force differential over 10,108 pairs** — the 28-value corpus plus 10 extra edges (`sha256:abcdef1234567890`, `V1.0.0`, `1.0.0+build.4/5`, `1.0.0-alpha/beta`, `1.0.00`, `1.0.0-`, `latest`) × all 7 package types — mapping the old `(comparable, newer)` pair onto the tri-state. **No answer moved.** It also modelled both call sites directly, old form vs new, over the same 10,108 pairs: identical.

It also caught something I would have had to think about: `compare_versions` drops the old `if current == latest: return False` short-circuit. That is absorbed by the branch structure — equal strings either both parse (`not_newer`) or neither does (`incomparable`, which the old pair scheme also treated as not-comparable) — and the differential confirms it empirically rather than by argument.

### I re-ran its differential against the post-fix head

Its review predates the two blocker fixes, one of which rewrote `is_version_orderable`. So its result did not strictly cover the current code. Re-running the same differential on `4d4bbce`:

```
differential over 10108 pairs (post-fix head)
  answers MOVED: 0
  is_version_orderable disagreements: 0
```

So the delegation rewrite preserved behaviour exactly, and the fable seat's conclusion holds for the head being merged.

### Its other findings, confirmed

- **No broken callers** — repo-wide sweep found the deleted names only in prose. `src/pmcp/manifest/__init__.py` never exported them; no `getattr` / re-export / `__all__` use.
- **Both call sites fail in the safe direction** — the short-circuit skips only on `"not_newer"`, so `incomparable` falls through to a refresh; `check_staleness` flags only on `"newer"`, so `incomparable` produces no fabricated notice.
- **No overclaim** in the CHANGELOG or the roadmap amendment.
- Rider commit `dde5168` checks out.

### Acted on its non-blocking note

`tests/test_refresher.py` described `is_version_newer` in the **present tense as live code** — a reader would go looking for a function this phase deleted. Reframed as history in `4d4bbce`, with the reason the test survives the migration made explicit. Left the other references alone deliberately: they are already past-tense and they are the record of why the tri-state exists.

Its second note — `is_version_orderable` now has zero `src/` callers — is now moot in the way that matters: after the EC-TRISTATE-1 fix it delegates to `compare_versions` rather than duplicating classification, so it is no longer a drift surface. Still test-only, still kept by design.

### Panel tally for this PR

4 seats: **3 AGREE, 1 DISAGREE**. The dissent produced two blocking findings, both real and both fixed. The strongest single contribution was a seat that agreed — the differential is now the verification of record for this change.

PR #172, opened 2026-08-23
feat: downstream catalog reconciliation and fan-out (FANOUT)

ViperJuice:
Executes **FANOUT**, phase 3 of `specs/phase-plans-v12.md`, per `plans/phase-plan-v12-FANOUT.md`. Closes v11's explicitly deferred item. **Draft — SL-2 and SL-3 still to land.**

## The problem

`subscriptions/listen` publishes the *gateway's own* catalog changes. It did **not** proxy a *downstream server's* `notifications/*` upward, so a downstream server adding a tool at runtime was invisible to subscribed clients.

Those notifications were reaching the gateway and being **dropped**: a JSON-RPC notification has no `id`, so it failed the `msg_id is not None and msg_id in managed.pending_requests` gate and fell through with no `else` branch — in **both** dispatch paths.

## Why this is reconciliation, not forwarding

The catalog is index-backed: `gateway.invoke` resolves via `get_tool` → `self._tools`, populated only by `_index_tools`. Forwarding a notification straight to the sink would tell a client "refetch" and hand it the **old** catalog, with a removed tool still invocable. The panel caught that at roadmap review.

So: a downstream `list_changed` schedules a per-server **re-index**, and the existing index mutators publish for free. Reconcile first, publish second.

## Three hazards, each verified before implementation

**Self-deadlock — immediate, not occasional.** `_index_capabilities` awaits `_send_request`, whose futures are resolved by the very read loop that received the notification (`set_result` at `:1791` / `:2010`). Reconciling inline makes the loop await a future only it can resolve. Reconciliation is `asyncio.create_task`, never awaited from dispatch, coalesced to one in-flight pass per server with a re-run flag.

**Storm on an unchanged catalog.** The `note_*` calls in `_index_*` and `_remove_server_indexes` are **unconditional**, so a naive remove-then-reindex publishes on *every* notification. The sink is suppressed during reconciliation via a depth counter, and the catalog compared before/after by **identifier sets** per kind — comparing counts would miss a rename.

**Two dispatch paths, not one.** `_handle_stdout_line` (stdio) and the SSE loop both wired. Both `TODO(post-P3B)` markers replaced so JSON-RPC `error` `code`/`data` survive to the `ClientManager` boundary.

Also handles a plan edge case: a downstream that announces a change then fails `tools/list` **rolls back** rather than leaving the catalog half-removed.

## An honest disclosure from the lane

SL-1 flagged that its rollback writes are a catalog mutation the existing publisher-coverage AST guard doesn't model — it doesn't see `dict.update`. I verified that: the guard exists, passes here, and genuinely has that blind spot. The lane correctly declined to widen a file it doesn't own. Worth a follow-up.

## Plan gap, recorded

SL-1's owned files were `manager.py` only, but it needed somewhere for its RED tests. I authorized `tests/test_client_manager.py` in the dispatch brief and it stayed contained to two clearly-named additions so SL-3 can work alongside. That is a gap in the plan's owned-files split, not lane drift — the docs lane will amend the roadmap.

## Verification so far

- `pytest -q` → **2604 passed, 3 skipped, 0 failed** (baseline 2585)
- Parent-tree leakage check clean; no file deletions

## Still to come

**SL-2** — two-transport emitter harness, including a new stdio fake server (today's harness is HTTP-only, so the stdio path has no emitter).
**SL-3** — behavioural tests asserting the *catalog* changed, on both transports.
**SL-docs** — CHANGELOG and roadmap amendments.

I will panel this before calling the phase done.

---
<a href="https://app.blacksmith.sh/Consiliency/codesmith/pmcp/pr/172"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790102139&installation_model_id=427051&pr_number=172&repository=Consiliency%2Fpmcp&return_to=https%3A%2F%2Fgithub.com%2FConsiliency%2Fpmcp%2Fpull%2F172&signature=e33ef1972dde33841eaf1f5ac8d106e922be793986c89865a07e5d0d159562cc"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,50 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- **A downstream server's own `notifications/tools/list_changed`,
+  `notifications/resources/list_changed`, and `notifications/prompts/list_changed`
+  now reach subscribed clients.** Previously the read loop parsed these frames
+  and silently dropped them on both transports — a notification has no `id`, so
+  it fell through the pending-request gate with no `else`. A downstream server
+  that added or removed a tool at runtime was invisible until the next
+  `gateway.refresh()`; that gap was v11 P3B's own Non-Goal.
+
+  This is reconciliation, not forwarding: `ClientManager`'s indexes back
+  `gateway.catalog_search`, `gateway.describe`, and `gateway.invoke`, so
+  relaying the raw notification to the subscription sink would have told a
+  client "refetch" and handed it the *old* catalog — with a tool the server
+  just removed still invocable. The gateway now re-indexes the announcing
+  server first and publishes only once that finishes, and only for the catalog
+  kinds whose identifier set actually changed (compared by identifier, not
+  count, so a rename still publishes). A downstream that announces a change
+  and then fails to re-list rolls its previous entries back rather than
+  leaving the catalog half-removed, and publishes nothing. **The guarantee for
+  a subscribed client:** the catalog is reconciled *before* the notification
+  goes out, so a client that refetches on receipt sees the change, every time.
+
+  Reconciliation runs as a spawned, per-server-coalesced background task
+  rather than inline in the read loop — re-indexing awaits a response that the
+  very read loop which received the notification is responsible for
+  resolving, so an inline await would deadlock the connection instantly. A
+  downstream that emits `list_changed` in reply to reconciliation's own
+  `tools/list` is bounded by a debounce on the re-run, not just coalescing, so
+  it costs one extra reconcile per interval instead of a hot spin. Both
+  transports are covered: stdio (`_handle_stdout_line`) and streamable
+  HTTP/SSE (`_read_sse`) previously shared the same silent-drop, and both now
+  dispatch through the same reconcile path. Unrecognised `notifications/*`
+  methods (progress, logging) remain a no-op, as before.
+
+### Fixed
+- **A downstream server's JSON-RPC `error` object no longer loses its `code`
+  and `data`.** Both dispatch paths kept only `message`, discarding the rest
+  of the `error` member. `ClientManager` now raises a typed `DownstreamError`
+  carrying `code` and `data` alongside `message`. Scoped to the
+  `ClientManager` boundary — `gateway.invoke` still maps every exception to
+  `E302` through `str(e)`, which is byte-identical to the old message, so no
+  `gateway.*` output changes; surfacing `code`/`data` to MCP clients is
+  tracked separately.
+
 ### Changed
 - **`version_checker.compare_versions(current, latest, package_type)` is now the
   sole version-classification path; `is_version_newer` and
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,10 +64,29 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   offers **zero** entries is a genuine answer — the server emptied that kind —
   and still clears the entries and publishes.
 
+  A reply the gateway cannot read is a failed listing too, and an **absent**
+  collection is not an empty one. A `tools/list` reply of `{}` — missing the
+  `tools` array the protocol requires — is malformed, not an announcement that
+  the server has no tools, and the same goes for a reply carrying something
+  other than an array in its place (`{"tools": {}}`, `{"tools": null}`). Each
+  of those now keeps the kind's previous entries and publishes nothing, exactly
+  like a request that failed; only a genuine array is an answer, and an empty
+  array still clears. Per kind, still: an unreadable `tools` reply no longer
+  costs an honest `resources` answer arriving in the same pass.
+
+  A catalog entry carrying no identity fails to parse rather than acquiring
+  one. A resource with no `uri`, or a prompt or tool with no `name` (or an
+  empty one), used to be indexed under a synthesized identifier of the form
+  `server::` — a catalog entry the downstream never offered, which replaced the
+  real entries and was published as a change. Such an entry is now skipped like
+  any other unparseable one, and a listing of nothing but those falls under the
+  all-unparseable rule above and keeps the previous entries.
+
   Failure classification is conservative by design. Any failure to list a kind
-  — a transport error, a server that does not implement it, or a listing that
-  could not be parsed at all — keeps that kind's previous entries; only an
-  explicit empty answer clears them. The accepted cost is the mirror case: a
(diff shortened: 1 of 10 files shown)


ViperJuice:
## Phase complete — three panel rounds, seven findings, all closed

Boarded **before** declaring the phase done this time, which is why nothing shipped broken.

### Round 1 — three findings, two blocking

- **D1 non-atomic mutation.** The catalog was emptied and *then* three network calls awaited, so a concurrent `gateway.invoke` got a spurious tool-not-found. Reproduced: `FOUND` → `NOT FOUND` mid-reconcile. This made the feature net-harmful — it could break invocations that previously worked.
- **D2 a kind whose re-list *failed* was published as removed.** All three seats found this independently. Exposure was **new**: on `main` these frames were dropped, so a connected server's resources could not vanish mid-session.
- **D3 content changes under a stable identifier were invisible.**

### Round 2 — two more

- **D5 a malformed entry wiped the server's catalog.** This one came from the repair lane **disclosing a residual rather than hiding it**. I reproduced it: one bad entry → `AttributeError`, catalog emptied, prior tool lost silently.
- **D6 nothing pinned the suppression unwind** — found by mutation, where deleting the decrement survived all 36 fanout tests.

### Round 3 — a seven-mutant battery, zero survivors, one real finding

- **F1 an all-malformed listing wiped the catalog and published the removal.** The per-entry guard handled *one* bad entry; when **every** entry failed to parse the kind still counted as answered. Reproduced: `['srv::alpha']` → `[]`, removal published.

  The decisive argument was that this **regressed against `bc68cbb`** — the deleted rollback preserved prior entries on any indexing exception — and that it contradicts the module's own principle at `:1610`, *"we could not ask is not they are gone."* Failing to parse an answer is the same epistemic state as failing to get one.

  Settled as **preserve-prior**, with the boundary intact. Verified all three cases: all-malformed preserves and publishes 0; legitimately empty still clears and publishes 1; one-bad-one-good indexes the good one.
- **F2 the CHANGELOG omitted round two**, including its deliberate connect-time change.

### Cleared, with reasons

**`-32601` is not being adopted.** The seat that originally proposed it ruled the conservative trade correct — *"deliberate, say so, ship it."* A server that legitimately empties a kind answers `[]` and is already handled; one that drops a capability mid-session is rare, and stale entries fail loudly at invoke. Documented as a chosen trade. The implementation is preserved as tag `salvage/fanout-32601` if it is ever wanted.

Also cleared: the apply block is await-free (proved empirically by inserting `await asyncio.sleep(0)` and watching the concurrent-read test catch it); the snapshot comparison cannot be stormed (the NaN candidate was disproven — CPython returns a cached singleton); coalescing loses no notification; a fetch resolving after a disconnect cannot apply zombie entries.

### Verification

- `pytest -q` → **2641 passed, 3 skipped, 0 failed** (baseline 2585)
- ruff, format, mypy clean; `test_publisher_coverage.py` passes untouched
- Roadmap validator 3 phases; docs-catalog 83 entries with `SPEC_COMPLIANCE.md` present
- **Every fix re-probed by me independently**, not taken from lane reports

One process note recorded against the follow-ups: I rotated lane worktrees while lanes were still live, which cost one lane its edits and killed another's test run. `refs/stash` is also repository-global across worktrees, so a concurrent pop elsewhere silently consumed a lane's stash. Both are avoidable with a signal-before-rotate convention and scratchpad copies instead of stash.

ViperJuice:
## Pre-merge advisor board on `6520cc9` — two blockers, both fixed in `84188f1`

Ran the board through the skill's own entry point (`compose_review_board` + `invoke_board`) rather than an ad-hoc script. It composed four lens-distinct seats: `grok-4.5` adversarial, `gpt-5.6-sol` red-team, `gemini-3.6-flash` alternative-approach, and `claude-fable-5` correctness — the last filled natively, since the TUI adapter path is for other hosts driving a Claude seat.

**Verdicts: 2 AGREE, 1 DISAGREE.** The red-team seat found two defects and I reproduced both before touching anything.

### G1 — an absent collection was read as an explicit empty one

`_fetch_server_listings` did `.get("tools", [])`, so a response *missing* its required field became `[]`:

```
result:{} -- tools key ABSENT      catalog=[]  published=1
tools:[]  -- explicitly empty      catalog=[]  published=1
```

Identical outcomes for opposite meanings — and a direct contradiction of the guarantee added **one commit earlier**: *"only an explicit empty answer clears them."* Absent is malformed; by this module's own principle that is not "they are gone".

### G2 — empty entry objects invented identities

`resource.get("uri", "")` turned `{}` into the identity `srv::`:

```
primed:      ['srv::mem://r1']
after [{}]:  ['srv::']   published: 1
```

A real entry replaced by a fabricated one. The MCP models reject those payloads; our own defaulting was manufacturing them.

Both now flow into machinery that already existed: an unreadable listing is a *failed* listing, and an entry without its identity fails to parse — hitting the per-entry skip, and if nothing parses, the all-malformed preserve-prior rule.

### The pattern was the real finding

This was the **fourth consecutive round** on the same class — a false removal reaching subscribed clients — with a different trigger each time: a failed listing, one malformed entry, an all-malformed listing, and now absent-collection plus empty-identity. Every fix was correct; every one left another path to the same outcome.

So the lane was told to assume a fifth exists and audit every path from a downstream response to a catalog write. Verified after the fix, alongside the two blockers and three non-regression cases:

| case | catalog | published |
|---|---|---|
| absent collection | preserved | 0 |
| explicit `[]` | cleared | 1 |
| all-malformed | preserved | 0 |
| one bad + one good | good indexed | 1 |
| `[{}]` entries | preserved, no `srv::` | 0 |
| collection is a dict | preserved | 0 |
| collection is `null` | preserved | 0 |
| entry is not an object | preserved | 0 |

### Verification

- `pytest -q` → **2655 passed, 3 skipped, 0 failed** (baseline 2585)
- ruff, format, mypy clean; `test_publisher_coverage.py` passes untouched
- Roadmap validator 3 phases; every case above re-probed by me, not taken from the lane's report

PR #174, opened 2026-08-23
fix: follow nextCursor when listing a downstream catalog (#173)

### LINKED ISSUE #173: Catalog listings are fetched one page deep; nextCursor is never followed
Found independently by two seats of the pre-merge advisor board on Consiliency/pmcp#172 (correctness and red-team). Neither fixed it, both for the same reason: it needs a semantics decision, not a repair. Filed rather than rushed into a fifth repair round.

## What happens

`ClientManager._fetch_server_listings` sends `tools/list`, `resources/list` and `prompts/list` with `params={}` and reads exactly one page. **There is no `nextCursor` handling anywhere on the catalog-listing path** — the only cursor machinery in `manager.py` is on the tasks path (`:3205`).

Reproduced against a downstream with 3 tools and a page size of 2:

```
downstream really has: a, b, c   (nextCursor = "p2")
gateway catalog:       ['srv::a', 'srv::b']
```

Tool `c` is invisible and uninvocable. Page 2 is never requested — `params` is `{}` on every call.

## What is and is not new

**The truncation is pre-existing.** `main` before Consiliency/pmcp#172 did the same: `tools/list` with `{}`, and connect already removed-then-indexed. A paginating server's catalog has always been page-one-only, at connect and at refresh.

**What Consiliency/pmcp#172 adds** is that reconciliation now runs on every downstream `notifications/*/list_changed` and **publishes a `list_changed` to subscribed clients asserting the catalog is current** — when it may be truncated. So the gateway now asserts a freshness it cannot back up, and re-applies the truncation on every notification rather than once at connect.

That is the same family as the defect Consiliency/pmcp#172 exists to fix — "we did not read all of it" surfacing as "the rest is gone" — but milder, and unlike the four defects fixed during that PR's review it fires on **ordinary operation**: no malformed response required, just a server with more entries than its page size.

## Why the obvious minimal fix is wrong

The tempting minimum — treat a reply carrying `nextCursor` as a failed listing, so prior entries are preserved and nothing is published — closes the false assertion but **freezes a paginating server's catalog permanently**: every listing carries a cursor, so every listing is "unreadable" and the catalog can never update. That is worse than the problem.

## What this actually needs

1. A real cursor loop for all three listing kinds, accumulating pages until no `nextCursor` is returned.
2. **A decision on what a mid-pagination failure means.** Both reviewers landed in the same place independently, and I agree: a failed page *N* should make the **whole** listing unreadable for that kind — preserve prior entries, publish nothing. Merging a partial result is precisely the false-removal shape the whole review series was about.
3. A bound on total pages or entries, so a misbehaving or malicious downstream cannot spin the loop forever. Note `LimitsPolicy.max_tools_per_server` already exists and interacts with this.
4. Tests: a multi-page listing assembles completely; a failure on page 2 preserves and publishes nothing; a cursor loop that never terminates is bounded.

Connect and refresh share `_fetch_server_listings`, so a fix reaches them too — which is the desirable outcome, but it means this is not reconcile-only and should be verified on all three paths.

## Adjacent findings from the same board round

Recorded here so they are not lost; all pre-existing and none blocking:

- **`max_tools_per_server` truncation is untested behaviourally.** A mutation of the truncation boundary (`>=` → `>`) survived the full `tests/test_client_manager.py`. Only its configurability is covered (`test_baseline_constraints.py:504`), never its effect.
- **`adopt_process` (`manager.py:2977`) indexes with no prior removal**, so the connect-path reasoning "connect removes the server's indexes first" does not cover it. Low stakes, install-time; worth a docstring amendment.
- **`LimitsPolicy.max_tools_per_server` has no `ge=1`** (`types.py:953`). At `0`, every non-empty listing classifies as "all entries unparseable" and logs a message saying so, which is untrue. Behaviourally inert today; `Field(ge=1)` would prevent the lie.
- **`tool.get("inputSchema", {})`** manufactures an accept-anything schema for a missing required field — the same manufacture-a-value pattern as the identity defaults fixed in Consiliency/pmcp#172, at lower stakes. Worth deciding deliberately rather than inheriting.
- **Duplicate identities within one listing collapse silently** (last write wins) while the indexed count counts each, so counts can overstate.

ViperJuice:
Closes Consiliency/pmcp#173.

Found independently by **two seats** of the pre-merge advisor board on Consiliency/pmcp#172 — the correctness and red-team lenses. Neither fixed it, both for the same reason: it needed a semantics decision rather than a repair.

## The defect

`tools/list`, `resources/list` and `prompts/list` were sent once with `params={}`, and whatever came back was taken as the entire catalog. There is no cursor handling anywhere on the listing path — the only cursor machinery in `manager.py` is on the tasks path.

Reproduced with a downstream holding 3 tools at page size 2:

```
downstream really has: a, b, c   (nextCursor = "p2")
gateway catalog:       ['srv::a', 'srv::b']    ← c invisible, page 2 never requested
```

**The truncation predates fan-out.** `main` sends `{}` too, at connect and at refresh. What made it worth fixing now is that reconciliation publishes `list_changed` *asserting the catalog is current* over a possibly-partial view, and re-applies the truncation on every downstream notification rather than once at connect.

## The semantics decision

**A failure on any page makes the whole kind unreadable, not partial** — prior entries kept, nothing published. Merging the pages that did arrive would drop entries the server still has and announce the drop: the same false-removal shape corrected four times during Consiliency/pmcp#172's review.

"Failure" reuses `_listing_entries` per page, so a page that isn't an object, is missing its collection, or carries a non-list collection is unreadable on exactly the same terms as page one.

**The tempting minimal fix was rejected** in the issue: treating any reply carrying `nextCursor` as a failed listing closes the false assertion but freezes a paginating server's catalog permanently, since every one of its pages carries a cursor.

## Bounds

`_MAX_LISTING_PAGES = 50`, sized against `max_tools_per_server`'s default of 100 and the tens-of-entries page sizes servers actually use. A repeated or self-referential cursor, a non-string cursor, and exceeding the cap all resolve to **unreadable** rather than to a truncated index — indexing a truncated view is the defect, so the bound must not resolve to "keep what we got".

Connect and refresh share `_fetch_server_listings`, so they read complete listings now too.

## Verification

- `pytest -q` → **2662 passed, 3 skipped, 0 failed** (baseline 2655)
- ruff, format, mypy clean; `test_publisher_coverage.py` passes untouched
- **6 of the 7 new tests are RED** against single-page behaviour. The seventh is the explicitly-empty-still-clears pin — a regression guard rather than a defect, correctly passing both ways.

Every invariant from the prior rounds re-verified: absent collection preserves; explicit `[]` clears and publishes; all-unparseable preserves; one-bad-one-good indexes the good one; dict/`null`/non-object shapes preserve.

## Not in scope

The five adjacent pre-existing findings recorded in Consiliency/pmcp#173's body — the untested `max_tools_per_server` truncation boundary, `adopt_process` indexing without prior removal, the missing `ge=1` on that limit, `inputSchema` defaulting, and silent duplicate-identity collapse — are unrelated to pagination and stay open. I'll split them into their own issue so Consiliency/pmcp#173 closes cleanly on its actual subject.

---
<a href="https://app.blacksmith.sh/Consiliency/codesmith/pmcp/pr/174"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790118848&installation_model_id=427051&pr_number=174&repository=Consiliency%2Fpmcp&return_to=https%3A%2F%2Fgithub.com%2FConsiliency%2Fpmcp%2Fpull%2F174&signature=55e7a2088a562fc7ca006ec2dba00df0e743bf7af5eeee28bd9dbfdbd90f45f6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   truncated view.
 
   Connect and refresh share this path, so they read complete listings too.
+  One consequence worth stating plainly: a server whose `tools/list` fails on a
+  *later* page now connects successfully with an empty tools catalog, where
+  before pagination existed there was no later page to fail. Only a page-one
+  failure is still a connect error. That matches how connect already treats a
+  first page whose every entry is unparseable, but such a server sits at zero
+  tools until a downstream notification or a refresh reconciles it.
 
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,27 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- **Catalog listings are now read to the end.** `tools/list`, `resources/list`
+  and `prompts/list` were sent once with no cursor and whatever came back was
+  taken as the whole catalog, so a downstream with more entries than its page
+  size had the rest silently missing. That truncation predated downstream
+  fan-out, but once reconciliation began publishing `list_changed` it started
+  asserting the catalog was current over a partial view — and re-applying the
+  truncation on every downstream notification rather than once at connect.
+  `nextCursor` (and the `next_cursor` spelling) is now followed to the end.
+
+  A failure on **any** page makes the whole kind unreadable rather than
+  partial: prior entries are kept and nothing is published. Merging the pages
+  that did arrive would drop entries the server still has and announce the drop
+  — the same false-removal shape this module has been corrected for repeatedly.
+  Following the cursor is bounded, and a server that repeats a cursor or never
+  stops paginating is treated as unreadable rather than looping or indexing a
+  truncated view.
+
+  Connect and refresh share this path, so they read complete listings too.
+
+
 ### Added
 - **A downstream server's own `notifications/tools/list_changed`,
   `notifications/resources/list_changed`, and `notifications/prompts/list_changed`
--- a/src/pmcp/client/manager.py
+++ b/src/pmcp/client/manager.py
@@ -95,12 +95,19 @@ _RECONCILE_RERUN_DEBOUNCE_S = 0.25
 # forever, and reconciliation runs on every downstream notification, so the
 # loop is reachable by a misbehaving peer rather than only at connect.
 #
-# Sized against the thing that consumes the result: `max_tools_per_server`
-# defaults to 100, and the MCP page size servers actually use is tens of
-# entries, so 50 pages clears any honest catalog by a wide margin while still
-# bounding a server that never stops paginating. Exceeding it makes the kind
-# UNREADABLE, not partial -- indexing a truncated view is what #173 was about.
-_MAX_LISTING_PAGES = 50
+# Exceeding it makes the kind UNREADABLE, not partial -- indexing a truncated
+# view is what #173 was about, so the bound must not resolve to "keep what we
+# got".
+#
+# That cuts both ways, which is why this is 500 and not the 50 first written
+# here (ah board review). Failing closed on an HONEST server is its own defect:
+# page size is the server's choice and 1 is legal, so a catalog of 100 tools --
+# `max_tools_per_server`'s own default -- can legitimately need 100 round trips,
+# and a 50-page cap would report that server as having no tools at all. The cap
+# is a runaway guard, not a catalog-size limit; 500 sequential round trips for
+# one kind is already pathological by any honest reading, while leaving real
+# small-page servers well inside it.
+_MAX_LISTING_PAGES = 500
 
 
 class DownstreamError(Exception):
--- a/src/pmcp/client/manager.py
+++ b/src/pmcp/client/manager.py
@@ -90,6 +90,18 @@ DOWNSTREAM_LIST_CHANGED_METHODS: dict[str, str] = {
 # the single pending re-run.
 _RECONCILE_RERUN_DEBOUNCE_S = 0.25
 
+# Ceiling on `nextCursor` follows for one catalog kind (Consiliency/pmcp#173).
+# A downstream that always returns a cursor would otherwise spin the fetch
+# forever, and reconciliation runs on every downstream notification, so the
+# loop is reachable by a misbehaving peer rather than only at connect.
+#
+# Sized against the thing that consumes the result: `max_tools_per_server`
+# defaults to 100, and the MCP page size servers actually use is tens of
+# entries, so 50 pages clears any honest catalog by a wide margin while still
+# bounding a server that never stops paginating. Exceeding it makes the kind
+# UNREADABLE, not partial -- indexing a truncated view is what #173 was about.
+_MAX_LISTING_PAGES = 50
+
 
 class DownstreamError(Exception):
     """A JSON-RPC `error` object returned by a downstream MCP server.
@@ -1675,17 +1687,15 @@ class ClientManager:
         """
         name = managed.config.name
 
-        tools_result = await self._send_request(managed, "tools/list", {})
+        tools_entries = await self._fetch_listing_pages(managed, "tools")
 
-        resources_task = self._send_request(managed, "resources/list", {})
-        prompts_task = self._send_request(managed, "prompts/list", {})
+        resources_task = self._fetch_listing_pages(managed, "resources")
+        prompts_task = self._fetch_listing_pages(managed, "prompts")
         listing_results = await asyncio.gather(
             resources_task, prompts_task, return_exceptions=True
         )
 
-        listings: dict[str, list[dict[str, Any]] | None] = {
-            "tools": _listing_entries(name, "tools", tools_result)
-        }
+        listings: dict[str, list[dict[str, Any]] | None] = {"tools": tools_entries}
         for kind, result in (
             ("resources", listing_results[0]),
             ("prompts", listing_results[1]),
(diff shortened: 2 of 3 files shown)


ViperJuice:
## Advisor board: 1 AGREE / 1 PARTIALLY AGREE / 1 DISAGREE — two blockers and one caveat, all fixed

This change was written on the main thread rather than by an isolated lane (a usage cap prevented dispatching one), so it had **no independent eyes until this board**. That is exactly what it caught.

### P1a — falsey cursors published a truncated catalog

Cursor absence was detected with `result.get("nextCursor") or result.get("next_cursor")` and `if not cursor`. So `nextCursor: 0` and `nextCursor: ""` read as "no more pages", page one was taken as the whole catalog, and both **slipped past the non-string rejection meant to catch them** — recreating the very truncation this PR exists to stop.

```
nextCursor=0   -> 1 entry  (page 2 never fetched)
nextCursor=""  -> 1 entry
```

Presence is now decided by **key**. An explicit `null` still ends the listing; a falsey or non-string cursor is unreadable.

### P1b — a late tools page aborted unrelated kinds

`tools` was awaited to completion before `resources` and `prompts` were started, so a failure on tools **page two** escaped `_fetch_server_listings` entirely and neither other kind was ever requested. Probed: only two `tools/list` calls happened, and a healthy resources update sat unapplied.

That contradicted my own docstring, which says only a page-**one** tools failure raises. All three kinds are now gathered together, and `_fetch_listing_pages` propagates only page one.

### Non-blocking caveat, also fixed — the cap failed closed on honest servers

The adversarial seat pointed out the cap cuts both ways. Making an over-paginating kind unreadable is correct; **failing closed on an honest server is its own defect.** Page size is the server's choice and 1 is legal, so 100 tools — `max_tools_per_server`'s own default — can legitimately need 100 round trips. At 50, that server was reported as having **no tools at all**.

Raised to 500 and reframed as a runaway guard rather than a catalog-size limit. Verified: page size 1 with 100 tools now assembles all 100.

### One test corrected, not weakened

`test_reconcile_is_spawned_and_not_awaited_in_the_dispatch_path` asserted `calls == ["tools/list"]`, which encoded the **old sequencing** — tools alone in flight because the other two had not been started. Gathering all three puts three requests in flight, so that equality no longer holds.

The property under test is unchanged and still carried by the two assertions that matter: the handler returned while the gate is shut, and nothing was indexed. The removed assertion was incidental to an ordering this PR deliberately changes. I have not claimed to mutation-prove it — an inline-await implementation *hangs* rather than failing, which is inherently hard to mutate-test.

### Verification

- `pytest -q` → **2668 passed, 3 skipped, 0 failed**
- ruff, format, mypy clean
- All four new board-fix tests proven RED against the pre-fix code
- Concurrency contracts re-probed: page-one `tools/list` failure still raises; an unsupported kind still returns `None` without costing the others; both paged kinds assemble concurrently

ViperJuice:
## Fourth seat: AGREE — 10/10 mutants killed, and it answered the question I was least sure of

The correctness seat reviewed at `b4dfb26` and re-verified at `0170a87`. Notably, it **independently found the falsey-cursor defect** before the other seats' fixes landed — its mutation battery at `b4dfb26` had two survivors, and it traced one to the live bug: `nextCursor: 0` and `""` slipping past `if not raw_cursor` and publishing a partial page. Re-run at `0170a87` with ten mutants: **10/10 killed**.

### On whether correcting the scheduler test was legitimate

**Yes**, with a sharper argument than I made. The property is carried by three things that all survive: the handler is a synchronous call, so reaching the next line at all proves dispatch returned; one assertion proves the reconcile started; two more prove it did not complete while the gate is shut.

The removed `calls == ["tools/list"]` additionally pinned that resources and prompts had **not yet been started** — which is exactly the tools-first sequencing this PR removes *because* it let a late tools page starve the other kinds. In its words: keeping it would have preserved a bug pin, not a guard. And equality-vs-membership makes no difference to the hang-not-fail mode, since an inline await never reaches the gate release either way.

### On a seventh trigger in the new concurrency shape

None found, with specifics: the later-page `except Exception` correctly does not catch `CancelledError`, so cancellation propagates rather than being converted to "unreadable"; the gather branches test `isinstance(result, BaseException)`, so a captured `CancelledError` cannot be mistyped as a listing; the page-one re-raise leaks no coroutines because `gather` awaits all three first; and cursor state is per-loop-instance, so three concurrent loops cannot cross-thread cursors.

### Two findings taken rather than deferred (`9febf5d`)

**A real mutant survived the battery.** Every pagination test was two pages deep, so shrinking `_MAX_LISTING_PAGES` to 2 passed the whole suite — nothing pinned that the cap is a runaway guard rather than a catalog-size limit. Added a ten-page assembly test and **verified it kills exactly that mutant**.

**An undocumented behaviour change.** A server whose `tools/list` fails on a *later* page now connects with an empty tools catalog; before pagination there was no later page to fail. Only page one is still a connect error. Consistent with how connect already treats an all-unparseable first page, but it is a contract change and such a server sits at zero tools until a notification or refresh pokes it. Now said plainly in the CHANGELOG.

Left alone as the seat suggested: the key-precedence corner where a server emits both cursor spellings with `nextCursor: null` plus a real `next_cursor`. `next_cursor` is not in the MCP spec and the shape requires a self-contradictory server.

### Final board tally for this PR

**4 seats: 2 AGREE, 1 PARTIALLY AGREE, 1 DISAGREE** — three sets of findings, all closed. Every one was in code I wrote on the main thread without lane review, which is why it was boarded.

`pytest -q` → **2669 passed, 3 skipped, 0 failed**; ruff, format, mypy clean.

PR #178, opened 2026-08-24
docs(plans): phase plan for v12 UPDPATH

ViperJuice:
The last v12 phase, planned. Three lanes: **SL-1** package-identity gate across all three refresh sites, **SL-2** ambient-environment contract pinned and documented, **SL-3** terminal docs sweep. SL-1 and SL-2 are disjoint and parallel.

All seven `EC-UPDPATH-*` criteria are referenced. `validate_plan_doc` and `validate_plan_dispatch_hints` are both clean (0 errors, 0 findings; one heuristic warning about a release evidence-reducer lane, explained in Execution Notes — this phase dispatches no release).

## What planning turned up

**A live bug the roadmap didn't know about.** `cli.py:770` computes `cache_path = get_cache_path(args.cache_dir)`, and the `--check-versions` branch at `:776` calls `check_staleness()` with no arguments. So `pmcp refresh --check-versions --cache-dir X` silently inspects the default cache. EC-6 requires a *CLI-level* regression test; without this wiring such a test would have to reach past the CLI to monkeypatch — the "unit test dressed as a CLI test" EC-6 exists to rule out. SL-1 owns the fix, and it's a user-visible change that earns its own CHANGELOG line rather than being smuggled in as a fixture detail.

**Acceptance criteria name test classes, not `-k` substrings.** The `-k` form was drafted first, then measured against the tree at v2.3.0:

| filter | collected today |
|---|---|
| `-k identity` | 29 / 2697 |
| `-k environ` | 8 / 2697 |
| `-k refresh_all` | 1 / 2697 |
| `-k unknown_package` | 1 / 2697 |
| `-k check_versions` | none |

Four of five would have been **green before any lane ran** — unfalsifiable criteria — and the fifth would have failed the command outright for collecting nothing. They now name four test classes the lanes are contracted to create, each verified to collect zero today, so each proving command fails now and can only pass when the work lands.

**Tests folded into their subject lanes.** The roadmap sketched A=`refresher.py`, B=`handlers.py`+docs, C=`tests/`. A standalone test lane can't satisfy "every impl task is preceded by a test task in the same lane" and would collide with both impl lanes on write ownership. Still three lanes; the deviation is recorded in Execution Notes rather than made silently.

**EC-4 is documentation, not behaviour — confirmed, not assumed.** Every stdio spawn reaches `sanitized_subprocess_env` (`env_store.py:124`), whose first statement is `os.environ.copy()`, evaluated at spawn time. `_connect_stdio` calls it at `manager.py:2151`. So connect, refresh, auto-reconnect and `update_server`'s restart already share one live-environment contract. SL-2.2 is docstring-only **by contract**, and the lane must stop and report if it turns into a behaviour change — freezing the ambient environment is an explicit Non-goal.

## The part most worth reviewing

SL-1's risk is **phrasing, not logic**. A guard written as "compare only when both sides are known" is the natural way to reach for it, passes a naive suite, and reintroduces the exact fail-open collapse that shipped three times (#155, #156, #163) before TRISTATE deleted the wrappers to make it unrepresentable. The plan requires SL-1.1 to contain a test that fails under *that phrasing specifically* — not merely one that passes under the correct one. If that requirement is too weak as written, it is the thing to say so about.

---
<a href="https://app.blacksmith.sh/Consiliency/codesmith/pmcp/pr/178"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790133607&installation_model_id=427051&pr_number=178&repository=Consiliency%2Fpmcp&return_to=https%3A%2F%2Fgithub.com%2FConsiliency%2Fpmcp%2Fpull%2F178&signature=8ec509a6764c4b68604222f139e5f2a47412ab4db5f5e5f4bb6d21e66a6141cc"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


ViperJuice:
## Board result — 4 seats: 1 AGREE, 1 DISAGREE, 1 UNAVAILABLE, 1 ERROR

| Seat | Lens | Verdict |
|---|---|---|
| grok-4.5 | adversarial | **AGREE** — confirmed every claimed fact against the tree, no blocking defects |
| gpt-5.6-sol | red-team | **DISAGREE** — three blocking findings |
| gemini-3.6-flash | alternative-approach | ERROR (the known headless-permission failure) |
| claude-fable-5 | correctness | filled natively; reporting separately |

All three red-team findings were **verified against source before acting on them**, and all three hold. Fixed in `d41b554`.

### 1. EC-UPDPATH-4 was factually wrong — the important one

The criterion asserted that "a credential rotated during the probe window is picked up by the restarted server." It is not. Confirmed empirically rather than by reading:

```
_refresh_config_unchanged: False
child envs differ       : True
=> update_server REFUSES: True
```

A manifest credential is resolved into `LocalMcpServerConfig.env` (`config/loader.py:1025`), so rotating it makes the recheck's child environment differ from the probe's, and `update_server` returns `ok=False` **without restarting** (`handlers.py:5090-5165`).

That refusal is the 2.2.1 TOCTOU guard behaving exactly as designed — *"the config restarted onto is the config that was probed."* So the roadmap's **decision** (document, don't freeze or refuse) stands untouched; what was wrong was its description of what there is to document.

EC-4 now has two halves: **4a** a config-driven change is refused and reported; **4b** a genuinely ambient variable is live at spawn, because both sides derive from one `stripped_base` and it cancels out. Worth noting what the old wording would have cost: a test written to it could only have passed by picking a variable that avoids the credential path — the same vacuous shape this plan had already removed once from its `-k` filters.

### 2. EC-UPDPATH-3 was insufficient

Routing the pair through `refresh_server` does not stop a wrong-package entry surviving. `refresh_all` returns the *existing* entry when `refresh_server` returns `None` or raises (`refresher.py:358-367`), and its final merge re-adds every cached entry missing from the new set (`:376-378`). New class `TestMismatchedCacheNeverSurvivesFailure` covers both paths.

### 3. The identity contract could not express identity

`GeneratedServerDescriptions` stores `package` as a bare name with no ecosystem (`types.py:1359-1367`), so npm `foo@1.0.0` against pypi `foo@1.0.0` reads as the same package — and unlike the docker case, these all produce orderable release versions, so `incomparable` never fires. My "cross-ecosystem is already handled" claim was true only for version-vs-digest.

`IF-0-UPDPATH-1`'s signature now carries both package **types**, and a new `IF-0-UPDPATH-2` adds `package_type` to the cached entry defaulting to `None`. An absent value reads as unknown, which EC-7 already forces to refresh — so the migration is safe by construction and needs no cache-format version bump.

### Also fixed

Grok's non-blocking note that `SL-1.5` should not race ahead of the refresher impl is addressed by the re-numbered task DAG. The roadmap's Key files gains `types.py` and `cli.py`. The plan is re-pinned to the amended roadmap sha.

`validate_plan_doc`, `validate_plan_dispatch_hints` and `validate-roadmap` all clean; the three new test classes confirmed to collect zero tests today, so each proving command fails now and can only pass when the work lands.

ViperJuice:
## Fourth seat in — board complete: 2 AGREE, 1 DISAGREE, 1 ERROR

`claude-fable-5` (correctness lens, filled natively) returned **AGREE** with **no must-fix-before-merge findings**, having re-derived every load-bearing claim from source rather than from the bundle's citations. Its execution guidance is folded in at `9954fe8`.

### The apparent conflict between seats, resolved

The correctness seat wrote that the single-`stripped_base` construction means "an ambient rotation during the probe window cannot trip a refusal — EC-4's 'no source change needed' is correct **by construction**." The red-team seat said the opposite. They are not actually in conflict:

- A **purely ambient** variable only reaches `stripped_base`, which is read **once**, post-probe. It cancels out. Correctness seat is right.
- A **manifest credential** is not purely ambient — `config/loader.py:1025` copies it into `LocalMcpServerConfig.env`, so it rides the `own` arm of `{**stripped_base, **own}`, and `own` **is** re-resolved at recheck. Red-team seat is right.

Confirmed by re-resolving a real manifest server across a rotation:

```
server: postgres | env_var: POSTGRES_URL
probe   config.env : OLD-TOKEN
recheck config.env : ROTATED-TOKEN
=> configs differ   : True
```

Each seat was right about its own case. The original one-sentence EC-4 conflated the two, which is precisely what the 4a/4b split separates — so the correction stands, and the correctness seat's "by construction" argument is now what backs **4b** specifically.

### Two findings that would have shipped wrong

**`check_staleness`'s polarity is inverted relative to `refresh_server`.** The former compares `== "newer"` to **add** to the stale dict (`:422`); the latter compares `== "not_newer"` to **skip** a refresh (`:221`). So an identity mismatch must *add* the server to `stale_servers`, not refuse a short-circuit — a lane porting the guard mechanically from `refresh_server` gets it backwards, and nothing in the plan said so.

**SL-2.1(b) had two vacuous-pass shapes.** A fully mocked `ClientManager` has no spawn to capture (`_connect_stdio` builds the env internally at `client/manager.py:2151`), so the test would be green regardless — it must patch at `asyncio.create_subprocess_exec` under a real `_connect_stdio`. And rotating a *managed* secret key exercises the `own` arm, which is 4a's path; 4b must rotate a genuinely non-managed key. Given this plan has already had to remove vacuous checks twice, that is the finding I am most glad arrived before execution.

### Also folded in

- The **configured-unknown arms of `_same_package` have no possible behavioural RED test** — `refresh_server` falls back to a command-string `pkg_name` that is never empty. Direct predicate assertions are the only pin, and SL-1.1 must include them.
- **RED proof is per-test, not per-command**, or a non-discriminating companion hides behind a red classmate.
- `tests/conftest.py` is owned by no lane — lanes define fixtures in-file or fail the closeout's dirty-path check.
- Existing SL-1-owned tests will need fixture updates (expected work, not regression), while `test_short_circuit_is_a_single_compare_versions_call` must keep passing — it is what catches a degenerate always-`False` gate.
- `docs/` does not exist; `README.md` is where the contract actually lands.
- Line precision: `cli.py:771` not `:770`; `client/manager.py:2151`.

### Confirmations worth recording

The seat independently reproduced the `-k` collection counts (29 / 8 / 1 / 1 / 0 of 2697), confirmed none of the named test classes exist, and verified the two claims I most expected to be stale: the docker cross-ecosystem case really is already handled — pinned by an existing `test_incomparable_pair_does_not_short_circuit` (`test_refresher.py:647`) — and neither half of EC-4 needs a source change.

Validators still clean.

Issue #183, opened 2026-08-25
update_server installs and executes a registry package derived from a misparsed command (npm run <script> → npx -y <script>@latest)


ViperJuice:
Found by the red-team seat of the advisor board reviewing the plan for #180, and confirmed on `main` @ `f5ea5ec` (v2.4.0).

This is a **different class** from the identity collisions in #180 and #182. Those mislabel a cache entry. This one causes pmcp to **install and execute a package from a public registry** whose name came from a misparse.

## What

`gateway.update_server` builds its update probe from the name `detect_package_type` returned (`src/pmcp/tools/handlers.py:5049`):

```python
if package_type == "npm":
    update_cmd = ["npx", "-y", f"{package_name}@latest", "--help"]
```

`npx -y` **installs the package if absent, without prompting**, then runs it.

The npm branch of `detect_package_type` has no subcommand handling, so the first non-flag token is taken as the package name:

```
detect_package_type("npm", ["run", "mcp"])    -> ("npm", "run")
detect_package_type("npm", ["exec", "old-pkg"]) -> ("npm", "exec")
detect_package_type("npm", ["create", "foo"])   -> ("npm", "create")
```

`npm run <script>` names a **local script in package.json** — not a registry package at all. So for a server configured as `npm run mcp`, calling `gateway.update_server` resolves the identity to `run`, then executes:

```
npx -y run@latest --help
```

That fetches and runs whatever `run` happens to be on the npm registry. The same holds for any first token the parser mistakes for a package name.

## Why it matters

- The executed name is **not a package the operator ever configured**. It is an artefact of a parsing bug.
- `npx -y` suppresses the install prompt, so there is no interactive gate.
- The name is attacker-*influenceable* in the ordinary dependency-confusion sense: a short generic token like `run`, `exec`, or `create` is exactly the kind of name someone can register on a public registry and wait for.

To be clear about severity, since it bounds the blast radius: reaching this requires a server configured with a form the parser mishandles **and** an operator invoking `gateway.update_server` on it. It is not remote-triggerable. But the failure mode is arbitrary package installation and execution, not a stale cache entry, which is why it is filed separately rather than folded into the parser work.

## Repro

```python
from pmcp.manifest.version_checker import detect_package_type
assert detect_package_type("npm", ["run", "mcp"]) == ("npm", "run")
# gateway.update_server would then run: npx -y run@latest --help
```

## Suggested scope

[...]

Issue #182, opened 2026-08-25
detect_package_type: value-carrying flags are not skipped, so uvx/pip/cargo collide the same way #180's forms do


ViperJuice:
Found by the correctness seat of the advisor board reviewing the plan for #180, and confirmed empirically on `main` @ `f5ea5ec` (v2.4.0).

#180 records collisions in the **docker** and **npm** branches. The same root cause is present in **uvx, pip and cargo**, and the #180 plan's research summary wrongly claimed those branches were "not implicated". They are:

```
uvx   ['--python', '3.12', 'pkg-a']                   -> ('pypi', '3.12')
uvx   ['--python', '3.12', 'pkg-b']                   -> ('pypi', '3.12')
uvx   ['--with', 'requests', 'srv-a']                 -> ('pypi', 'requests')
uvx   ['--with', 'requests', 'srv-b']                 -> ('pypi', 'requests')
pip   ['install', '--index-url', 'https://x', 'pkg-a'] -> ('pypi', 'https://x')
pip   ['install', '--index-url', 'https://x', 'pkg-b'] -> ('pypi', 'https://x')
cargo ['run', '--features', 'full', 'srv-a']          -> ('cargo', 'full')
cargo ['run', '--features', 'full', 'srv-b']          -> ('cargo', 'full')
```

## Root cause

Those branches skip **flags** (`arg.startswith("-")`) but not the **values those flags carry** (`src/pmcp/manifest/version_checker.py:284-314`). So the first non-flag token is the flag's *value*, not the package.

The docker branch already solves exactly this, with a `_value_flags` set enumerating flags whose next token must be skipped (`version_checker.py:336-359`). uvx, pip and cargo have no equivalent.

Note `uvx --from thing other -> ('pypi','thing')` is **correct** — but only by luck: `--from`'s value happens to *be* the package. That single safe flag is what made the branch look fine.

## Why it matters

Same consequence as #180. v2.4.0's identity gate compares these names to decide whether a cached description still describes the configured package, so two different servers launched with the same flag value read as the **same package** and `_same_package` confirms identity across a real swap.

**`uvx` is the most consequential of all of them** — it is a common MCP launcher and `--with` is a normal pattern, which makes this more realistic than the `registry:host:port` case #180 is built around.

## Why not folded into #180

[...]

Issue #180, opened 2026-08-24
detect_package_type collapses distinct packages for some command forms, so the identity gate can falsely confirm


ViperJuice:
Found by the red-team seat of the pre-merge board on Consiliency/pmcp#179 (v12 UPDPATH) and confirmed empirically.

`detect_package_type` returns a package **name** that is not unique for several legal command forms, so two genuinely different packages resolve to the same identity:

```
docker ['run', 'registry:5000/old-image'] -> ('docker', 'registry')
docker ['run', 'registry:5000/new-image'] -> ('docker', 'registry')
npm   ['exec', 'old-pkg']                 -> ('npm', 'exec')
npm   ['exec', 'new-pkg']                 -> ('npm', 'exec')
```

The docker case comes from `raw.split(":")[0]` (`src/pmcp/manifest/version_checker.py:316`), which is parsing a registry `host:port` as an image tag separator. The npm case takes the first argument as the package name without accounting for `exec`.

## Why it matters now

This is **pre-existing** — the parser has always behaved this way, and before UPDPATH nothing compared package names, so a collapsed name only ever produced a wrong *version lookup*. UPDPATH makes it consequential: `_same_package` compares these names to decide whether a cached description still describes the configured package. For the forms above it **confirms identity across a real package swap**, so the freshness short-circuit fires and the wrong package's tool descriptions keep being served — exactly the defect UPDPATH exists to close, unmet for those command forms.

So EC-UPDPATH-1, -3 and -6 hold for the ordinary forms (`npx -y <pkg>`, `uvx <pkg>`, `docker run <image>`) and do **not** hold for these.

## Why it was not fixed in UPDPATH

`detect_package_type` backs every version lookup in the codebase, not just the identity gate. Changing its parsing has blast radius well beyond that phase and deserves its own RED tests and review round rather than being appended to a closeout. Deliberately deferred by the maintainer with the phase's limitation recorded in `specs/phase-plans-v12.md`.

## Repro

```python
from pmcp.manifest.version_checker import detect_package_type
assert detect_package_type("docker", ["run", "registry:5000/old-image"]) == ("docker", "registry")
assert detect_package_type("docker", ["run", "registry:5000/new-image"]) == ("docker", "registry")
```

## Suggested scope

- Parse a docker reference properly: a `:` only separates a tag when it appears **after** the last `/`, so `registry:5000/img` has no tag while `img:1.2` does.
- Skip an `exec` / `run` subcommand before taking the npm package name.
[...]

PR #184, opened 2026-08-25
fix(parser): reconcile docker reference splitting and skip npm subcommands (narrows #180)

ViperJuice:
Narrows **#180**. Deliberately does **not** close it — see *What this does not fix*.

2.4.0's identity gate decides whether a cached description still describes the configured package by comparing the *name* `detect_package_type` returns. A name that is stable across two **different** packages is therefore not a cosmetic parsing wart: `_same_package` reads it as a positive confirmation, and the freshness short-circuit goes on serving the wrong package's tool descriptions.

## Two causes, both closed

**Docker split on the first `:`** — so `registry:5000/old-image` and `registry:5000/new-image` both resolved to `registry`, the registry host rather than an image. A colon only introduces a tag in the final path segment; before the last `/` it is a `host:port`.

The correct rule already existed in the same module as `_docker_image_tag`, which is what pin detection uses. So the fix adds its **paired complement** `_docker_image_name` — same design as `_strip_npm_tag`/`_npm_tag` — rather than a second, divergent implementation. My first draft did write a second rule, and the board caught that it both regressed `img:1.2@sha256:abc` (→ `img:1.2@sha256`, where today's code correctly gives `img`) and disagreed with `_docker_image_tag` on that same input.

**npm subcommands were taken as the package** — `npm exec old-pkg` and `npm exec new-pkg` both resolved to `exec`. A leading subcommand is now skipped **once**, and for `npm` only:

- *once*, not repeatedly, because `i` and `exec` are real npm packages — a docker-style anywhere-loop turns `npm install i` and `npm exec exec` into `None`
- *npm only*, because `npx -y exec` genuinely names a package called `exec`
- `x` is included; without it `npm x` collides exactly as `npm exec` did

`_npm_package_arg` gains a **required** `command` parameter, not a defaulted one: a default would let one of the two shared call sites keep old behaviour silently while type-checking clean — the exact drift the shared scan exists to prevent.

## Two live defects found along the way

Neither is in #180; both fell out of the same shared-scan bugs.

- **A docker digest was reported as a version pin.** `img@sha256:abc` reported a pin of `abc` — a fragment of the digest — and `img:1.2@sha256:abc` reported `1.2@sha256:abc` rather than `1.2`.
- **A pin on an `npm exec` server was reported as unpinned.** `npm exec pkg@1.2` scanned to `exec`, which carries no version suffix, so a real pin was missed.

## What this does **not** fix

My plan asserted that docker's image-token *selection* was sound and only tag-stripping was broken. **The board falsified that.** Identity is still collapsed wherever a flag's *value* is taken as the package name — and these are ordinary forms:

```
docker run --env-file .env old-image / new-image     -> both ('docker', '.env')
docker run --mount type=bind,src=/a A / B            -> both ('docker', 'type=bind,...')
npm exec --package=old-pkg / new-pkg -- shared-bin   -> both ('npm', 'exec')
```

Tracked on **#182** (broadened for this), with **#183** for the more serious consequence: `update_server` builds `npx -y {parsed}@latest --help`, and `npx -y` installs without prompting, so a misparsed `npm run <script>` installs and executes a registry package.

## Verification

`2717 passed, 3 skipped, 0 failed`; ruff clean; mypy clean (42 files).

**Six mutants, each run against one test node alone** — never the class. A class-level run reddens on any member, so it cannot show the specific test is load-bearing; that is the hole two hollow tests shipped through in this repo's last two phases.

| mutant | kills |
|---|---|
| docker → `raw.split(":")[0]` | the docker collision test |
| `skip_subcommand = False` | both npm collision tests |
| skip fires repeatedly | both real-package-name tests |
| digest not stripped first | **only** the 2 digest cases, 8 others green |
| pin site → `_npm_package_arg(args, "npx")` | the pin matrix |
| `_docker_image_tag` stops stripping digest | the pin matrix |

The last two are what pin the helpers as complements: each half has its own mutant, so a divergence cannot pass silently.

Collision tests assert **inequality** of real pairs rather than single expected values — a value assertion passes against a parser that collapses both forms onto some other shared name.

## Migration

Affected servers refresh **once**: a docker server on a `host:port` registry or an `npm exec` server now has a different package identity than its cache recorded, so that entry fails the identity check once and regenerates. Same one-time migration 2.4.0's `package_type` addition caused.

---
<a href="https://app.blacksmith.sh/Consiliency/codesmith/pmcp/pr/184"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790282235&installation_model_id=427051&pr_number=184&repository=Consiliency%2Fpmcp&return_to=https%3A%2F%2Fgithub.com%2FConsiliency%2Fpmcp%2Fpull%2F184&signature=71242e0ca44f025c6faedc90ad3135677acede1c20ab320e953bd462d9f04d67"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/.consiliency/evidence/mutation-180.md
+++ b/.consiliency/evidence/mutation-180.md
@@ -1,0 +1,122 @@
+# Mutation evidence — Consiliency/pmcp#180
+
+Every proof runs **one test node alone** against the mutant, never the class.
+A class-level run reddens on any member, so it cannot show that the specific
+test is load-bearing — the hole through which two hollow tests shipped in this
+repo's last two phases (board finding, adversarial + red-team seats).
+
+## 1. docker branch reverted to `raw.split(":")[0]`
+
+    FAILED ...::TestPackageIdentityCollisions::test_docker_registry_host_port_distinguishes_images
+    1 failed in 0.08s
+
+## 2. npm subcommand skip disabled (`skip_subcommand = False`)
+
+    FAILED ...::TestPackageIdentityCollisions::test_npm_exec_distinguishes_packages
+    1 failed in 0.10s
+    FAILED ...::TestPackageIdentityCollisions::test_npm_x_alias_distinguishes_packages
+    1 failed in 0.07s
+
+## 3. skip fires repeatedly (docker-style anywhere-loop)
+
+    FAILED ...::TestNpmSubcommandSkipFiresOnce::test_npm_install_of_a_package_named_i
+    1 failed in 0.09s
+    FAILED ...::TestNpmSubcommandSkipFiresOnce::test_npm_exec_of_a_package_named_exec
+    1 failed in 0.09s
+
+## 4. digest NOT stripped first (the first draft's rule)
+
+    FAILED ...::TestDockerReferenceSplitting::test_name_and_tag_agree[img@sha256:abc-img-None]
+    FAILED ...::TestDockerReferenceSplitting::test_name_and_tag_agree[registry:5000/img@sha256:abc-registry:5000/img-None]
+    2 failed, 8 passed in 0.11s
+
+Note this mutant fails **only** the digest cases and leaves the other eight
+green — the parametrised ids name exactly which forms the ordering protects.
+
+## 5. pin-detection call site reverted to `_npm_package_arg(args, "npx")`
+
+    FAILED ...::TestUpdateServerVersionRepair::test_detect_effective_version_pin_matrix
+    1 failed, 185 deselected in 0.95s
+
+## 6. `_docker_image_tag` stops stripping the digest
+
+    FAILED ...::TestUpdateServerVersionRepair::test_detect_effective_version_pin_matrix
+    1 failed, 185 deselected in 1.02s
+
+Mutants 5 and 6 are what pin the two helpers as complements: each half of the
+pair has its own mutant, so they cannot drift apart silently.
+
+## 7. pin check ignores the digest (the regression this PR introduced and then fixed)
+
+Removing the `_docker_image_digest` consultation from
+`_detect_effective_version_pin`:
+
+    FAILED ...::TestUpdateServerVersionRepair::test_detect_effective_version_pin_matrix
+    1 failed, 185 deselected in 1.11s
+
+This mutant is the defect the red-team seat found in the first implementation.
+Stripping `@digest` before the tag scan was correct for the NAME, but reading
+the tag alone then reported a digest-only reference as unpinned -- so a
+digest-pinned server, the tightest pin docker has, read as unpinned and
+`update_server` would have pulled `image:latest`, restarted the unchanged
+config, and recorded the registry's newest digest as though it had updated.
+
+Before this PR the same reference produced the garbage pin `'abc'`, which was
+wrong but at least truthy, so the refusal still fired. The first draft of the
+fix turned a wrong-but-safe value into a right-looking-but-unsafe `None` --
+a reminder that "more correct in isolation" is not the same as "safe in
+composition".
+
+## 8. flag-skip and subcommand check swapped (adversarial seat's surviving mutant)
+
+    FAILED ...::TestNpmSubcommandSkipFiresOnce::test_a_leading_flag_does_not_consume_the_subcommand_skip
+    1 failed in 0.08s
+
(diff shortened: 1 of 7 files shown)


ViperJuice:
## Second board round on the implementation — all four seats, four findings, all closed in `360f226`

| Seat | Verdict |
|---|---|
| grok-4.6 (adversarial) | DISAGREE — surviving mutant A |
| gpt-5.6-sol (red-team) | DISAGREE — digest-pin regression |
| claude-fable-5 (correctness) | PARTIALLY AGREE — confirmed A, found B |

### 1. A regression this PR introduced (red-team, confirmed by correctness)

Stripping `@digest` before the tag read was correct for the **name** and broke the **pin**:

```
BEFORE  img@sha256:abc  -> pin 'abc'   (garbage, but truthy → refusal fired)
AFTER   img@sha256:abc  -> pin None    (reads as UNPINNED)
```

A digest is the tightest pin docker offers. With `None`, `update_server` would pull `image:latest`, restart the unchanged digest-pinned configuration, and record the registry's newest digest — **announcing an update while still running the old immutable image**. I converted wrong-but-safe into right-looking-but-unsafe, which is strictly worse than the bug I was fixing.

Fixed with `_docker_image_digest`, the third member of the reference-splitting family, consulted **before** the tag. `img:latest@sha256:…` is the same trap wearing a tag that the `== "latest"` test would otherwise discard. All eight docker pin forms verified.

### 2. Surviving mutant A (grok and fable, independently)

Swapping the flag-skip and subcommand checks passed **all 2717 tests**. The code was already correct — `npm --silent exec old-pkg` → `old-pkg` — but nothing *pinned* the ordering, because every npm test put the subcommand first. Correct by accident is not correct by contract. Now pinned; the mutant dies.

### 3. Surviving mutant B (fable, unique) — the guard I wrote was itself hollow

`detect("npm","npx",["-y","exec@1.2"]) == "1.2"`, commented *"npx still must not skip a package genuinely named exec"*, **cannot enforce that**. The skip matches the raw token against `_NPM_SUBCOMMANDS`, and `exec@1.2` is not in the set — so a call site hardcoded to `"npm"` returns the identical `1.2`:

```
correct : npx -y exec@1.2     -> '1.2'
mutant  : npx -y exec@1.2     -> '1.2'    <- no divergence, guard useless
correct : npx -y exec pkg@1.2 -> None
mutant  : npx -y exec pkg@1.2 -> '1.2'    <- divergence
```

That is the **third hollow test in three phases**, and the second written by the agent that had just documented the failure mode. Reading an assertion cannot distinguish a hollow one from a real one; only mutating the exact line it claims to pin can.

### 4. Tooling: stale bytecode fabricates false red *and* false green

A full-suite run reported mutant A's new test as FAILING against correct source. Cause: a pure-reorder mutant has identical byte size, and apply/compile/restore inside one second defeats `.pyc` invalidation (whole-second mtime + size). Purging `src/**/__pycache__` → 5 passed.

The same mechanism can fabricate a false **green** — certifying a mutant as killed when it was not. All nine proofs re-confirmed with `__pycache__` purged and `PYTHONDONTWRITEBYTECODE=1`, and `.consiliency/evidence/mutation-180.md` now requires it for any future proof in this repo.

### Evidence location

Moved from scratch into the repo at `.consiliency/evidence/mutation-180.md`. The plan required the mutation output *recorded*; scratch is not the repo, so that acceptance criterion was genuinely unmet until now (red-team finding 2).

### Verification

`2718 passed, 3 skipped, 0 failed` with pycache purged and bytecode disabled; ruff clean; mypy clean. CI green on all ten checks.

PR #185, opened 2026-08-25
fix(update): refuse to probe when no package identity is recoverable (closes #183)

### LINKED ISSUE #183: update_server installs and executes a registry package derived from a misparsed command (npm run <script> → npx -y <script>@latest)

ViperJuice:
Closes **#183**.

## The defect

`gateway.update_server` builds its update probe from the parsed package name — `npx -y {name}@latest --help` — and **`npx -y` installs without prompting**.

A server configured as `npm run mcp` names a **script** in the local `package.json`, not a registry package. The parser returned it as one, so pmcp fetched and executed whatever occupied that name on the public npm registry. Short generic script names — `run`, `start`, `dev`, `mcp` — are exactly the kind someone can register and wait on.

Reaching this needed a server configured with an affected form **and** an operator invoking `update_server` on it; it was not remotely triggerable. But the failure mode is arbitrary package installation and execution, not a mislabelled cache entry.

## The fix is in the parser, not at the execution site

`handlers.py:5040` **already** refuses cleanly on `("unknown", None)`, before any probe is constructed. The defect was that `npm run mcp` never reached that guard, because the parser handed back a plausible-looking name.

So `npm run <script>` and `npm create <initializer>` now report **no recoverable package identity**, and the existing refusal does the rest — the same rule the identity gate follows: *cannot confirm, so do not act on a guess*. Adding a second guard at the probe site would have been a divergent check for the same property, and this repo has spent three phases paying for divergent implementations of one rule.

**`npm create` was wrong in a second way the issue didn't record:** npm resolves `npm create foo` to the package **`create-foo`**, so returning `foo` named a genuinely different package than the one npm would run.

## Scoped to subcommands, not to names

| refuses | unchanged |
|---|---|
| `npm run mcp` | `npx -y run` → `run` |
| `npm run start` | `npx -y create` → `create` |
| `npm create foo` | `npm exec run` → `run` |
| `npm --silent run mcp` | all 8 other real-package forms |

`run` is a real npm package; nothing here should make `npx -y run` unresolvable. The refusal keys on the *subcommand position*, not on the token.

## Verification

`2722 passed, 3 skipped, 0 failed`; ruff clean; mypy clean — with `__pycache__` purged and `PYTHONDONTWRITEBYTECODE=1`, per the tooling note in `.consiliency/evidence/mutation-180.md` (stale bytecode fabricated both a false red and, potentially, a false green during #184).

**The end-to-end test asserts no probe was executed**, not merely that the parse changed — the parse being right only matters if it reaches the refusal. Mutation-proved: removing the refusal fails that test.

```
=== MUTANT: script-name refusal removed (the #183 defect) ===
FAILED tests/test_tools.py::...::test_update_server_never_probes_a_script_name
1 failed in 2.06s
```

---
<a href="https://app.blacksmith.sh/Consiliency/codesmith/pmcp/pr/185"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790288414&installation_model_id=427051&pr_number=185&repository=Consiliency%2Fpmcp&return_to=https%3A%2F%2Fgithub.com%2FConsiliency%2Fpmcp%2Fpull%2F185&signature=1a2821211192d3e0db9c5a62906954cde936c3063109a3716d0912e9ff4c5c42"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/.consiliency/evidence/mutation-183.md
+++ b/.consiliency/evidence/mutation-183.md
@@ -1,0 +1,63 @@
+# Mutation evidence — Consiliency/pmcp#183
+
+Protocol: `src/**/__pycache__` purged and `PYTHONDONTWRITEBYTECODE=1` on every
+apply/restore, per the tooling note in `mutation-180.md`. Stale bytecode
+fabricated a false RED during #184 and can equally fabricate a false GREEN.
+
+## 1. The refusal removed (the #183 defect itself)
+
+    FAILED ...::TestUpdateServerVersionRepair::test_update_server_never_probes_a_script_name
+    1 failed in 2.06s
+
+The test asserts **no probe was executed**, not merely that the parse changed.
+The correctness seat independently reproduced this end-to-end on a copy by
+reverting `version_checker.py` to main, and captured what the defect actually
+produces:
+
+    UpdateServerOutput(ok=True, ...,
+      message="Updated and restarted 'scripted' (npm:mcp); the new version is now active.")
+
+Fabricated success, plus `npx -y mcp@latest` installed and run from the public
+registry. That is the whole issue in one line of output.
+
+## 2. Allowlist reverted to the denylist that shipped first
+
+    FAILED ...::test_a_subcommand_without_a_package_operand_fails_closed[args9]
+    FAILED ...::test_a_subcommand_without_a_package_operand_fails_closed[args10]
+    5 failed, 9 passed, 136 deselected in 0.33s
+
+**This is the mutant that mattered, and the first implementation did not kill
+it.** The original fix denied `run` and `create` only, so everything else fell
+through as an installable package name:
+
+    npm start        -> ('npm', 'start')      -> npx -y start@latest
+    npm test         -> ('npm', 'test')       -> npx -y test@latest
+    npm run-script X -> ('npm', 'run-script') -> npx -y run-script@latest
+    npm init foo     -> ('npm', 'init')       -> npx -y init@latest
+    npm rum mcp      -> ('npm', 'rum')        -> npx -y rum@latest   (a TYPO)
+
+Every one of those names resolves on registry.npmjs.org. All three CLI seats
+plus the correctness seat found this independently and all four recommended the
+same repair: invert to an allowlist.
+
+The design lesson is the reusable part. A denylist **fails open**, which is
+backwards for a fix whose purpose is preventing unintended execution: the cost
+of omitting an entry is arbitrary package execution, while the cost of an
+over-broad allowlist is only that an unusual launch form cannot be
+auto-updated. I had closed the two forms the issue named and left the class
+open -- treating the issue's examples as the specification.
+
+Two related errors worth recording:
+
+* `run-script` is npm's **canonical** command; `run` is the alias. I fixed the
+  alias and missed the real name.
+* `init` is the alias of `create`, and I fixed only `create`.
+
+## Pre-existing tests inverted, not weakened
+
+`test_npm_command` and `test_npm_with_no_subcommand_still_finds_the_package`
+both asserted `npm -y server-pkg` -> `server-pkg`. That is not a legal npm
+invocation (npm requires a subcommand), so those assertions had encoded the
+parser's old permissiveness -- the general form of this very defect. They are
+now inverted, with the reason in their docstrings, rather than the guard being
+relaxed to keep them green.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,28 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Security
+- **`gateway.update_server` no longer installs and executes a registry package
+  derived from a misparsed command.** The update probe is built from the parsed
+  package name — `npx -y {name}@latest --help` — and `npx -y` installs without
+  prompting. A server configured as `npm run mcp` names a **script** in the
+  local `package.json`, not a registry package, but the parser returned it as
+  one, so pmcp fetched and ran whatever occupied that name on the public
+  registry. Short generic script names (`run`, `start`, `dev`, `mcp`) are
+  exactly the kind that can be registered and waited on.
+
+  `npm run <script>` and `npm create <initializer>` now report **no recoverable
+  package identity** rather than a wrong one, and `update_server` refuses on
+  that before constructing any probe — the same rule the identity gate follows:
+  cannot confirm, so do not act on a guess. (`npm create foo` was wrong in a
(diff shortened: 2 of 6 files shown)


ViperJuice:
## Board result — all four seats found the same defect, all four recommended the same repair

| Seat | Verdict |
|---|---|
| grok-4.6 (adversarial) | DISAGREE |
| gemini-3.7-flash (alternative-approach) | DISAGREE |

Fixed in `b1ae257`.

### The defect: my fix was a denylist, and a denylist fails open

I denied `run` and `create` — **the two forms the issue named** — and everything else still fell through as an installable package name:

```
npm start        -> ('npm','start')      -> npx -y start@latest
npm test         -> ('npm','test')       -> npx -y test@latest
npm run-script X -> ('npm','run-script') -> npx -y run-script@latest
npm init foo     -> ('npm','init')       -> npx -y init@latest
npm rum mcp      -> ('npm','rum')        -> npx -y rum@latest      (a typo)
```

Every one of those names resolves on registry.npmjs.org. For a fix whose entire purpose is preventing unintended package execution, default-allow is backwards: omitting an entry costs **arbitrary package execution**, while an over-broad allowlist costs only the ability to auto-update an unusual launch form.

Two related errors the same mistake produced, both caught by the correctness seat checking `npm help` rather than recalling: **`run-script` is npm's canonical command and `run` is the alias** — I fixed the alias and missed the real name — and **`init` is the alias of `create`**, of which I fixed only `create`.

### The repair

An allowlist: only `exec`, `x`, `install`, `i`, `add`, `dlx` put a registry package in the next position. Everything else — including typos and subcommands npm may add later — fails closed and hits the existing refusal before any probe is constructed.

**14 dangerous forms verified refused** (incl. `t`, `ci`, `tst`, `stop`, `restart`, flag-prefixed); **12 real-package forms unchanged**, including `npx -y run` and `npm exec run`. `pnpm`/`yarn` were already refused.

### Two pre-existing tests inverted, not weakened

Both asserted `npm -y server-pkg` → `server-pkg`. That is not a legal npm invocation — npm requires a subcommand — so those assertions had encoded the parser's old permissiveness, which is the general form of this defect. Inverted, with the reason in their docstrings, rather than relaxing the guard to keep them green.

### Also taken

The refusal message now names the command line: *"could not determine a registry package to update for 'x' from `npm run mcp`"*. It previously said *"could not determine package manager … Supported managers: npm (npx)"*, which reads as a contradiction when your command literally is npm. Pinned by an assertion.

### What the correctness seat verified that I could not assert myself

It reverted the parser **on a copy** and reproduced the defect end-to-end, confirming the test is not hollow:

```
UpdateServerOutput(ok=True, ...,
  message="Updated and restarted 'scripted' (npm:mcp); the new version is now active.")
```

Fabricated success plus `npx -y mcp@latest` fetched and run from the public registry.

It also found an **undocumented bonus fix**: before this, an `npm run mcp` server compared cache freshness against registry package `mcp`'s *real* version and could short-circuit "up to date" on an entirely different package.

### Verification

`2737 passed, 3 skipped, 0 failed`; ruff clean; mypy clean — `__pycache__` purged with `PYTHONDONTWRITEBYTECODE=1` throughout, per the tooling note from #184. Evidence in `.consiliency/evidence/mutation-183.md`.

PR #188, opened 2026-08-26
fix(ci): give every workflow job a timeout (closes #187)

### LINKED ISSUE #187: release.yml has no timeout-minutes, so a stuck job hangs for 6 hours and blocks publish
Hit while cutting v2.4.1.

## What happened

The release workflow's `build` job hung for **3h11m** on `Run tests with coverage`. Prior cuts (v2.2.1, v2.3.0, v2.4.0) all completed in roughly 4–5 minutes.

The tagged commit was healthy — running the exact release command locally on `4f81744` gave `2737 passed, 3 skipped`, coverage 85.86% against a 60% gate. So this was a runner/network stall, not the code.

Cancelling and re-running the same run cleared it.

## The gap

`.github/workflows/release.yml` sets **no `timeout-minutes`** on any job, so a stuck job runs until GitHub's default 6-hour ceiling. Because `publish` and `github-release` both `needs:` the previous job, a single stall silently blocks the entire release for six hours before anything fails.

`pipeline-bootstrap.yml:74` is the only workflow in the repo that sets a timeout (`30`). `test.yml` has none either, though the blast radius there is smaller — a stuck PR check is visible and doesn't gate a publish.

## Why it matters more here than elsewhere

The tag push **is** the publish: it builds, ships to PyPI via trusted publishing, and creates the GitHub release. A silent 6-hour hang on that path means a release appears tagged but never actually ships, and nothing surfaces the problem — I only noticed because I compared the run's start time against previous cuts.

## Suggested scope

- Add `timeout-minutes` to each job in `release.yml`. The build job's real work is ~5 min, so something like 20–25 leaves generous headroom while failing fast on a stall.
- Consider the same for `test.yml`.
- Consider adding `workflow_dispatch` to `release.yml`. Today the only trigger is a tag push, so recovering from a failed release run means either `gh run rerun` (worked here) or deleting and re-pushing a published tag, which rewrites a ref that may already have been fetched.

## Not urgent

The 6-hour ceiling is a backstop that does eventually fire, and `gh run rerun` recovers without touching the tag. Filing so the fix gets made deliberately rather than appended untested to a release that is mid-flight — CI plumbing that has never been exercised is exactly what failed here.

ViperJuice:
Closes **#187**. CI-only: 12 additive lines across four workflow files, no `src/` change.

## Why

Cutting v2.4.1, `release.yml`'s `build` job hung **3h11m** on "Run tests with coverage" against a healthy commit — the same command locally gave `2737 passed`, 85.86% coverage vs a 60% gate. Cancel + `gh run rerun` cleared it: a transient runner stall.

No job in `release.yml`, `test.yml`, `docker.yml` or `maintenance.yml` set `timeout-minutes`, so a stall ran to GitHub's 6-hour default. `publish` and `github-release` each `needs:` the job before them, so **one stall silently blocked the whole publish** — and since the tag push *is* the publish, a version can sit tagged-but-unshipped with nothing surfacing it. I only noticed by comparing the run's start time against previous cuts.

## Values — measured, not estimated

From `GET /actions/runs/{id}/jobs`, `completed_at - started_at`, last 5 runs. Roughly 4× observed p100 with a 10-minute floor.

| file | job | observed | timeout |
|---|---|---|---|
| `release.yml` | `build` | 3.6–4.4m | 20 |
| `release.yml` | `publish` | ~20s | 10 |
| `release.yml` | `github-release` | ~8s | 10 |
| `test.yml` | `test` ×3 | **5.30m p100** | **25** |
| `test.yml` | 5 other jobs | <1m | 10 |
| `docker.yml` | `build` | <1m | 20 |
| `maintenance.yml` | `notify-worker` | 9s | 10 |

`pipeline-bootstrap.yml` already had `timeout-minutes: 30` and is untouched.

The goal is failing fast on a *stall*, not policing variance — a job that legitimately grows 3× still passes. The board specifically noted that tightening the floor to 5 would be the real false-kill risk, for `min-version-smoke`.

### `workflow_dispatch` was proposed and **struck**

An earlier draft added it with a `tag` input, for recovering a stalled release without re-pushing a published tag. Two seats independently found it wasn't just a security widening — it was **incorrect**. `actions/checkout` pins no ref (`release.yml:12`, `:63`) and both tag derivations read `GITHUB_REF_NAME` (`:73`, `:97`), which under dispatch is the *selected ref*, not a workflow input. Dispatching `tag=v2.4.1` from `main` would publish **main's** artifacts under that version number.

Worse: **PyPI can succeed while the GitHub release fails**, leaving a half-published version. Both mitigations I'd offered fail on inspection — verified live, there are **no tag rulesets** (`gh api .../rulesets` → 0) and `environment: release` has `protection_rules: []`, so it gates nothing today.

Recovery already exists and is what unblocked v2.4.1: `gh run rerun`. And the timeouts are themselves the recovery improvement — a 6-hour silent hang becomes a ≤25-minute red X.

### Three acceptance-test bypasses, each reproduced

My original check asserted only that `timeout-minutes` was *present*. Proofs in `.consiliency/evidence/bypass-proofs-187.md`:

| bypass | original | hardened |
|---|---|---|
| `.yaml` file escapes a `*.yml` glob | exit 0 | exit 1 |
| `timeout-minutes: 360` — **re-creates the 6h default** | exit 0 | exit 1 |
| over-indented job silently swallowed | exit 0 | exit 1 |

The third is the dangerous one: over-indenting `publish:` makes it parse as a key inside `build`'s mapping. YAML is valid, `jobs` becomes `[build, github-release]`, and the presence check passes because every *surviving* job has a timeout. My plan claimed "the real proof is CI itself" — **false for `release.yml`**, which triggers only on tag push and never appears in PR checks. It would have surfaced at the next tag push, as a broken release.

The hardened check globs both extensions, validates the value (`int`, not `bool`, 1–30), skips `uses:` jobs (a reusable-workflow call can't carry a timeout), and compares job-name sets against `main`, **failing** rather than printing.

`actionlint` is now mandatory rather than best-effort — `release.yml` can't be exercised pre-merge, so it's the only Actions-schema check on the release path before a tag is cut. My original invocation was also fail-open (`&& actionlint || echo skipped` swallows a non-zero exit).

## Verification

- hardened acceptance check: **exit 0**; all three bypasses **exit 1**
- job sets identical to `main` across all five workflows
- `actionlint 1.7.12`: **exit 0**, no findings
- `pipeline-bootstrap.yml`: `git diff main` empty, still 30
- suite: `2737 passed, 3 skipped, 0 failed`

**Reviewer check the plan couldn't do pre-push:** this PR's check list should match a `main` PR's — same names, same count. `release.yml`'s jobs won't appear (tag-triggered), which is precisely why `actionlint` is load-bearing here.

## CHANGELOG

Intentionally untouched, stated rather than silently skipped: CI-only, no `src/` change, and the `changelog` job requires an entry only for `src/` changes.

---
<a href="https://app.blacksmith.sh/Consiliency/codesmith/pmcp/pr/188"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790308314&installation_model_id=427051&pr_number=188&repository=Consiliency%2Fpmcp&return_to=https%3A%2F%2Fgithub.com%2FConsiliency%2Fpmcp%2Fpull%2F188&signature=011e4884ac7c0cb6fedd03d59d8845f5b215b3433cec2b0fc02120b038e9955a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/.consiliency/evidence/bypass-proofs-187.md
+++ b/.consiliency/evidence/bypass-proofs-187.md
@@ -1,0 +1,63 @@
+# Acceptance-test bypass proofs — Consiliency/pmcp#187
+
+The first version of this plan's acceptance test had three bypasses, all found
+by the advisor board's correctness seat and all reproduced empirically before
+being acted on. Each is recorded with the original (exit 0, wrongly passing)
+and hardened (exit 1) result.
+
+The general shape: **a check that only asserts presence proves almost nothing.**
+This is the fifth round in this repo where a check I wrote looked adequate and
+was not, three of those written immediately after documenting that exact
+failure mode.
+
+## A. `.yaml` extension escapes the glob
+
+Original globbed `*.yml` only. GitHub runs both extensions.
+
+    original : missing: none                          EXIT=0   <- WRONG
+    hardened : ['sneaky.yaml:unbounded -> None']      EXIT=1
+
+## B. Presence checked, value not
+
+`timeout-minutes: "not a number"` passed. So did `timeout-minutes: 360` --
+which **re-creates the six-hour default this issue exists to fix** while
+satisfying the test.
+
+    original : missing: none                          EXIT=0   <- WRONG
+    hardened : ['release.yml:build -> 360']           EXIT=1
+
+Hardened form asserts `isinstance(v, int) and not isinstance(v, bool) and
+1 <= v <= 30`, and skips jobs with `uses:` (a reusable-workflow call cannot
+carry `timeout-minutes`, so requiring one there would force invalid YAML).
+
+## C. A job silently deleted still passes
+
+Over-indenting `publish:` makes it parse as a key *inside* `build`'s mapping.
+The YAML is valid, `jobs` becomes `[build, github-release]`, and the presence
+check passes because every *surviving* job has a timeout. The publish job is
+simply gone.
+
+    original : jobs [build, github-release], missing: none   EXIT=0   <- WRONG
+    hardened : release.yml: [build, github-release, publish]
+                         -> [build, github-release]          EXIT=1
+
+**This is the one that mattered.** The plan claimed "the real proof is CI
+itself" -- false for `release.yml`, which triggers only on tag push and never
+appears in PR checks. A mangled release.yml would have surfaced at the next tag
+push, as a broken release. The hardened check compares job-name sets against
+`main` and **fails** rather than printing.
+
+## Also corrected
+
+Live `test (3.10)` p100 is **5.30m** (measured over 5 runs), not the 4-5m the
+plan's table recorded. The plan's own acceptance criterion ("each value >= 4x
+observed") was therefore falsified by its own value of 20 on day one. Raised to
+25.
+
+## Why actionlint is mandatory here, not optional
+
+`release.yml` cannot be exercised pre-merge -- it triggers only on tag push --
+so actionlint is the *only* Actions-schema validation on the release path
+before a tag is cut. The plan's original invocation was also fail-open
+(`command -v actionlint && actionlint ... || echo skipped` swallows a non-zero
+actionlint into the `||` branch), which is the same defect class as B above.
--- a/.consiliency/evidence/bypass-proofs-187.md
+++ b/.consiliency/evidence/bypass-proofs-187.md
@@ -122,3 +122,74 @@ Check #1 skips jobs carrying `uses:` (a reusable-workflow call cannot take
 the exemption is currently untested against real input. It is correct to have --
 without it, a future `uses:` job would force invalid YAML to satisfy the check --
 but it would silently exempt any future job that gains a `uses:` key.
+
+---
+
+## Implementation-board round: check 0, and seven mutants the others miss
+
+Four seats on the code. grok AGREE, gemini AGREE, codex DISAGREE, fable
+PARTIALLY AGREE. Two findings were live; both are now closed.
+
+### The plan shipped the PROVEN-WEAK verification block
+
+Both dissenting seats found the same thing, and it is the sharpest finding of
+the round: this PR's *plan* still carried check #1 as `glob("*.yml")`
+presence-only, check #2 as print-only, and `actionlint` with the extension gap
+— **the exact three bypasses this evidence file proves are exploitable**, sitting
+one file away. The PR would have shipped the weak script as the canonical
+runnable procedure while the proof of its weakness travelled beside it.
+
+Fixed: the plan's Verification block is replaced with the hardened forms, with a
+note forbidding reintroduction of the weak variants, and its stale table row
+(`test` → 20, "4–5m") corrected to 25 / 5.35m p100.
+
+### Check 0 — audit the diff SHAPE — is stronger than 1–3 combined
+
+The correctness seat's contribution, and it changes what this change is guarded
+by. Checks 1–3 cover exactly *{job-name set} × {timeout presence and value}*.
+Every other byte of `release.yml` is unguarded. Seven mutants pass all three:
+
+| mutant | consequence |
+|---|---|
+| `environment: release` dropped | silent; publish runs unprotected |
+| `needs: publish` → `needs: build` | silent; GH release no longer gated on PyPI |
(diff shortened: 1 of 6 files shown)


ViperJuice:
## Implementation board — 2 AGREE, 1 DISAGREE, 1 PARTIALLY AGREE; both live findings closed in `ebe677c`

| Seat | Verdict |
|---|---|

### 1. The PR contradicted itself — the plan shipped the proven-weak checks

Both dissenting seats found this independently, and it's the sharpest finding of the round. The committed **plan** still carried check #1 as `glob("*.yml")` presence-only, check #2 as print-only, and `actionlint` with the extension gap — **the exact three bypasses the adjacent evidence file proves are exploitable**.

In a repo whose documented recurring failure is *"a committed check looked adequate and wasn't"*, this PR would have shipped the weak script as the canonical runnable procedure with the proof of its weakness sitting one file away. Fixed: hardened forms, a note forbidding reintroduction, and the stale table row (`test` → 20, "4–5m") corrected to 25 / 5.35m p100.

### 2. Check 0 — audit the diff *shape* — is stronger than checks 1–3 combined

The correctness seat's contribution, and it changes what this change is actually guarded by. Checks 1–3 cover exactly *{job-name set} × {timeout presence and value}*. **Every other byte of `release.yml` is unguarded.** Seven mutants pass all three:

| mutant | consequence |
|---|---|
| `environment: release` dropped | silent — publish runs unprotected |
| `needs: publish` → `needs: build` | silent — GH release no longer gated on PyPI |
| tag pattern `"v*"` → `"V*"` | **silent forever** — no run, so no red X ever |
| `on:` tags filter deleted | fires on every branch push |
| publish `needs: build` dropped | loud, but next-tag-only |
| `build` 20 → 3 (below p100) | job killed at next tag push |
| `maintenance.yml` deleted | nothing notices |

The tag-pattern case is #187's own nightmare in purer form: corrupt the pattern and **nothing runs at all**, so unlike a broken job there's never a red X.

Verified here — check 0 (`every changed line matches ^\+\s+timeout-minutes: \d+$, nothing removed`) **catches all three** silent `release.yml` mutants that checks 1–3 miss:

```
drop-environment -> check0 CATCHES (exit 1)
needs-build      -> check0 CATCHES (exit 1)
tag-case         -> check0 CATCHES (exit 1)
```

### 3. Floor raised 1 → 10

Red-team found that accepting any int in 1–30 leaves a live mutant: `build` 20 → 1 passes the predicate, preserves the job set, and passes `actionlint`, while killing a 4m10s job after one minute. Verified. The floor now matches the plan's own stated sizing rule, and every delivered value is ≥10.

```
build 20 -> 1, floor 10:  ['release.yml:build -> 1']  exit 1
```

### 4. Red-team's third finding — CI evidence — independently satisfied

It couldn't reach GitHub from its harness. Confirmed here: **11/11 green**, and the check-name set is **identical** to a `main` PR's (diffed against #186), so no job was silently disabled. Live durations under the new caps: `test` legs 3–4m against 25, everything else <1m against 10.

The correctness seat independently re-measured every value at 8-run p100 and confirmed all are ≥4.5×, and that the 20→25 raise was **necessary**, not conservative: 20/5.35 = 3.7× would have violated the plan's own criterion on day one.

### Filed, not fixed: #189

None of these checks runs *in CI* — they were executed by hand against this diff. A future PR touching `release.yml` inherits no protection, on the one file that never appears in PR checks. Filed as **#189** with the full mutant matrix rather than appended untested here.

PR #192, opened 2026-08-26
fix(parser): fail closed on unrecognised npm flags (closes #180)

### LINKED ISSUE #180: detect_package_type collapses distinct packages for some command forms, so the identity gate can falsely confirm

ViperJuice:
Closes **#180**, the last fail-open ecosystem. #182 converted uvx, pip, cargo and docker; npm was deliberately deferred because it needed its flag arity enumerated, and that enumeration is this change's real content.

## The residual

`_npm_package_arg` skipped anything starting `-` and took the next bare token, so a flag's **value** became the package name. Two different servers then shared one identity, which v2.4.0's gate reads as a **positive confirmation**. All five reproduced on 2.5.0; all five now resolve correctly:

```
npm exec --loglevel silly a/b        ('npm','silly')      -> ('npm','a') / ('npm','b')
npm exec --registry https://r a/b    ('npm','https://r')  -> ('npm','a') / ('npm','b')
npx -y --registry https://r a/b      ('npm','https://r')  -> ('npm','a') / ('npm','b')
npm exec --global false a/b          ('npm','false')      -> ('npm','a') / ('npm','b')
npm exec --color always a/b          ('npm','always')     -> ('unknown', None)  [refuses]
```

The last two are the ones a naive design misses. A board seat **implemented an earlier version of this plan faithfully in a copy** and demonstrated its acceptance suite passed while `--color always` still collided — the suite was blind to the exact entry the method got wrong.

## Source: `@npmcli/config`, not `npm config list --json`

The table is generated from npm's own `definitions` (181 entries with declared `type`) and `shorthands` (40 expansions). The generator is committed at `.consiliency/notes/derive_npm_flags.py` with a `--verify` mode.

`npm config list --json` was the original plan's source and is **disqualified**: it emits the active merged config, omits private keys, adds host-specific ones, and `NPM_CONFIG_COLOR=always` flips `color` from boolean to string — the same generator would classify differently on different hosts.

**Arity cannot be inferred from a default value.** `--color` has a boolean default but a declared type of `always|Boolean`, and `npm exec --color always` is legal. That falsified the original model outright.

## Classification, verified against npm's own parser

`null` is dropped before classifying — it means *unset*, not a value type. Without that, `yes`/`optional`/`production`/`workspaces` read as conditional and `--yes`/`-y`, a real MCP launch form, gets refused.

The conditional-arity set is exactly **`{browser, color}`**. A board seat proposed `{browser, color, proxy}`; `proxy` has no Boolean member and **always** consumes — including `--proxy a`, where it swallows the package — so it is a value flag, and grouping it as conditional would have refused an unambiguous form.

Shorthands expand with arity from expansion length: ≥2 bakes in a value ⇒ consumes nothing; length 1 is a rename ⇒ inherits the target's arity. This retires the previous hand-coded `-y` special case.

## Two self-corrections found during implementation

- **`npm exec -- <pkg>` was refused by the new fail-closed default.** A legitimate form the plan never named.
- **Baked-value shorthands had to split from real booleans.** `--silent` bakes in its value so it consumes nothing, leaving `npx --silent true arg` → `true`; `--global` is a real boolean that *does* swallow a literal `true`. Getting that backwards mis-parses both.

## Verification

```
2802 passed, 3 skipped, 0 failed      ruff + mypy clean
generator --verify: 0 mismatches over 181 definitions, cross-checked against nopt
manifest: 98 servers, 0 unknown
9-form residual sweep: none still colliding
```

Must-not-regress, all intact: `npm --silent exec pkg`→`pkg`, `npm exec pkg`→`pkg`, `npm install i`→`i`, `npm exec exec`→`exec`, `npx -y exec`→`exec`, `npx -y pkg`→`pkg`.

## Cost

A server launched with a flag npm's own schema does not describe can no longer be auto-updated — it fails closed rather than guessing. That is the deliberate trade this repo has now applied to all five ecosystems: failing closed is loud and recoverable, failing open is silent.

---
<a href="https://app.blacksmith.sh/Consiliency/codesmith/pmcp/pr/192"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790365226&installation_model_id=427051&pr_number=192&repository=Consiliency%2Fpmcp&return_to=https%3A%2F%2Fgithub.com%2FConsiliency%2Fpmcp%2Fpull%2F192&signature=f7e8cf9a021e1971b4e0e8d9a73cda07f43f9973bc985e7ea5bd5ca81c4c426c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/.consiliency/notes/derive_npm_flags.py
+++ b/.consiliency/notes/derive_npm_flags.py
@@ -1,0 +1,356 @@
+"""Derive npm's flag-arity tables from npm's OWN config schema, and verify them.
+
+`_npm_package_arg` must know, for every npm flag, whether the token after it is
+that flag's value or the package name. Getting one entry wrong in the
+"consumes nothing" direction is fail-OPEN: the flag's value becomes the package
+identity, and two different servers collapse into one (Consiliency/pmcp#180).
+npm has 181 flags and 40 shorthands, which is far past what anyone can recall
+correctly -- #182 removed ten memory-drafted entries across four ecosystems for
+exactly that reason. So the table is generated, never hand-listed.
+
+THE SOURCE IS `@npmcli/config/lib/definitions/index.js`, NOT `npm config list`.
+-----------------------------------------------------------------------------
+`npm config list --json` is disqualified as a source, and this is not a style
+preference. It emits the *active merged configuration*, not the definition set:
+
+  * it omits private entries (`_auth`, `password`),
+  * it adds host-specific ones (`npm-version`, a user-scoped registry),
+  * `NPM_CONFIG_COLOR=always` flips `color` from a boolean to a string,
+  * it omits all 40 shorthands wholesale, so `--silent` looks like an unknown
+    flag rather than an alias of `--loglevel silent`.
+
+The same generator would therefore classify differently on different hosts. The
+definitions module, by contrast, declares each flag's `type` -- the real arity
+declaration, independent of environment and of any default value.
+
+WHAT `type` MEANS, AND THE TWO MEMBERS THAT ARE NOT VALUE TYPES
+--------------------------------------------------------------
+`type` is a SET, e.g. `color` is `['always', Boolean]`. Two members are
+multiplicity/absence markers rather than value types and are dropped before
+classifying:
+
+  `null`   "unset". Dropping it is load-bearing: `yes`, `optional`,
+           `production`, `workspaces` and `expect-results` are all
+           `null|Boolean`, so without the drop they read as conditional and
+           `--yes`/`-y` -- a real MCP launch form -- would be refused.
+  `Array`  "may repeat". Measured inert on npm 11.19.0 (no definition changes
+           class either way, and none is `Boolean|Array`); the generator
+           reports it if that ever stops being true.
+
+Classification, after the drop:
+
+  Boolean alone         -> boolean-arity: consumes nothing ... except a literal
+                           `true`/`false`, which npm's parser DOES take as the
+                           flag's value. Verified below for all 76.
+  no Boolean member     -> value: consumes the next token. This is where
+                           `proxy` (`null|false|{url}`) belongs -- it has no
+                           Boolean member and always consumes, including
+                           `--proxy a` where it swallows the package.
+  Boolean AND non-Bool  -> CONDITIONAL: arity depends on the next token's
+                           content, so no single class is right. Left UNLISTED,
+                           which makes the scanner refuse. Fail closed.
+
+SHORTHANDS EXPAND; ARITY COMES FROM THE EXPANSION'S LENGTH
+----------------------------------------------------------
+  length >= 2  a value is baked in (`silent` -> ['--loglevel','silent'])
+               => boolean-arity, consumes nothing.
+  length == 1  a pure rename (`reg` -> ['--registry'], `y` -> ['--yes'])
+               => the target's arity.
+
+This subsumes the hand-coded `-y` special case `_npm_package_arg` used to
+carry, and makes `npm --silent exec pkg` -> `pkg` hold by construction.
+
+Deliberately NOT a pytest: npm's schema is version-specific and CI has no npm.
+Behaviour is pinned by the pure unit tests in tests/test_version_checker.py;
+this script is what keeps those pins honest against a real npm.
+
+Usage:
+    python3 .consiliency/notes/derive_npm_flags.py            # emit the tables
+    python3 .consiliency/notes/derive_npm_flags.py --verify   # check committed
+"""
+
+import json
+import pathlib
+import subprocess
+import sys
+
+REPO = pathlib.Path(__file__).resolve().parents[2]
+
+# Multiplicity/absence markers, not value types. See the module docstring --
+# dropping `null` is what keeps `--yes` usable.
+_NOT_A_VALUE_TYPE = frozenset({"null", "Array"})
+
+BOOLEAN, VALUE, CONDITIONAL, UNINTERPRETABLE = (
+    "boolean",
+    "value",
+    "conditional",
+    "uninterpretable",
+)
+
+# `--package` is handled as a KNOWN-POSITIVE by the scanner (its value IS the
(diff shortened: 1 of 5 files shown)


ViperJuice:
## Board — 2 AGREE, 1 DISAGREE; the dissent was right twice. Fixed in `429436c`

| Seat | Verdict |
|---|---|
| claude-fable-5 (correctness) | in progress |

Both live findings verified against npm's own parser before acting, and mutation-proved after.

### 1. Nullable booleans swallowed the package

The generator correctly drops `null` from a type list — it means *unset*, not a value type. But the **scanner** must still consume a literal `null` token after a nullable boolean, exactly as it does `true`/`false`:

```
npm exec --yes null A  ->  ('npm','null')
npm exec --yes null B  ->  ('npm','null')     COLLIDES
```

`_same_package` confirms that false identity. Affects `--yes`, `--optional`, `--production`, `--workspaces`, `--expect-results`. Now fixed — `null` joins the boolean-literal set.

Worth naming: my own correction to the plan-round finding ("drop `null`, it means unset") was right about **classification** and I failed to carry it through to **parsing**. The same word means different things in the two places.

### 2. An attached baked-value shorthand is not its spaced form

`--silent=true X` is not `--silent X`. npm expands the shorthand and the **attached** value becomes a positional, so nopt leaves `["true","X"]` — the package is `true`, not `X`. pmcp read `X` for both spellings, collapsing two genuinely different packages:

```
npm exec --silent=true  X   ->  ('npm','true')    [was ('npm','X')]
npm exec --silent=false X   ->  ('npm','false')   [was ('npm','X')]
```

The spaced form is deliberately unchanged — that's the whole distinction: `--silent` bakes its value so consumes nothing, while `--global` is a real boolean that swallows a literal `true`.

### 3. The third finding did not reproduce — but its cause is real

`derive_npm_flags.py --verify` is green here and does not report `--local-address`. The cause is worth recording: npm enumerates that flag's type from **the host's own network interfaces**, so its member list differs per machine and the committed table is not byte-reproducible across hosts. The seat simply had different interfaces.

Not a defect in this change, but the verifier should treat host-enumerated types as *non-comparable* rather than as drift. Filing separately rather than patching it here.

### Mutation proofs — per node, on a copy, provenance asserted

```
drop null from the literal set     -> the nullable-boolean test fails
attached baked value returns None  -> the attached-shorthand test fails
```

### Verification at `429436c`

```
2805 passed, 3 skipped, 0 failed     ruff + mypy clean
manifest: 98 servers, 0 unknown
18-form residual sweep: no real collisions
```

One sweep entry (`--silent=true X`) *looked* like a collision and is not — checked against nopt, both invocations genuinely name the package `true`, so pmcp agreeing is correct. **My sweep was varying a token npm does not treat as the package.** Recording it because a residual sweep that reports false positives is exactly as misleading as one that misses real ones.

### #180

It could **not** have closed on the previous commit — `--yes null` was a live collision. On this commit the class is closed as far as 18 measured forms reach.

